PR #68, opened 2022-12-15
Use Inter var font

rwharrisjr:


alucigna (approved):
Added the sample I used to validate/compare to our old Inter font usage and google's Inter font display tool (https://fonts.google.com/specimen/Inter).  

The "Inter var" version looked the best between the options.

PR #85, opened 2023-01-09
Add basic truncation to table header cells

### LINKED ISSUE #81: Add truncation to table headers and cells

Yes - when table columns become too small to fit the horizontal content, the column text gets cut off.

Truncation to automatically be added to table headers and cells.

We've tried adding our own truncation to every table's headers and cells, but it's causing issues when the `th` has a sort icon next to it. `ADataTable` should provide truncation out of the box so everyone who uses it doesn't have to remember to add this to every table.

Our attempt at trying to add this to all tables ourselves. Notice the issue with the `th` with the sort icon:

Screenshot 2023-01-05 at 4 59 30 PM

How the table looks with none of our custom truncation:

Screenshot 2023-01-05 at 5 05 07 PM

rwharrisjr:
Resolves #81 

This PR allows for passing style to the table header cell. Using this the header text will truncate if needed.

Given the nature of `<table>` and css, this does not make the table automatically truncate and size columns for responsive layouts. Other table implementations will need to be used for responsive design.

It is recommended to allow a table to horizontal scroll rather than truncate columns header text.

### DIFF
--- a/framework/components/ADataTable/ADataTable.js
+++ b/framework/components/ADataTable/ADataTable.js
@@ -4,6 +4,7 @@ import React, {forwardRef, useMemo, useRef, useCallback, useState} from "react";
 import AInView from "../AInView";
 import AIcon from "../AIcon";
 import ASimpleTable from "../ASimpleTable";
+import ATooltip from "../ATooltip";
 import "./ADataTable.scss";
 
 /**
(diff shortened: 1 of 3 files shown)


brennarvo:
This looks good from a visual standpoint, but my only concern is that there is no way to preview the full name of the column. For instance, I don't think there is an easy way to pass a `title` attribute to the column header given the nature of the data structure for header columns. I'm actually starting to think having that key/value pair specifically for table header row props is starting to make sense.

I'm a bit hesitant to approve, just because I feel like previewing the full text is important for a table header.

rwharrisjr:
@brennarvo I struggled with trying to get a tooltip working in there, and I couldn't seem to get it working consistently with the way the refs work, and the content is also an issue. Might be easier after adding the wrapper. Do you think a key/value for `tooltipLabel` or something to that effect would help here?

brennarvo (approved):
Forgot that `ADataTable` allows you to pass your own component for the header cell. If you set the headers with a width, then you can also pass your own tooltip or such in there to render the rest of the text.

PR #89, opened 2023-01-10
Data table sort alignment

rwharrisjr:


framework/components/ADataTable/ADataTable.js
```diff
@@ -71,7 +71,7 @@ TableCell.displayName = "TableCell";
 
 export const getSortIconName = (column, sort) => {
   if (!sort || column.key !== sort.key) {
-    return "sortEmpty";
```
brennarvo, line 74: I want to say that this condition should render the muted sort empty icon where both arrows are light gray. I added the sortEmpty icon in https://github.com/cisco-sbg-ui/magna-react/pull/67, but it looks like I didn't do it correctly (this was before I knew about the icon builder script).

Can we confirm that we should be using the default sort icon instead of the muted sort empty?

rwharrisjr:
@brennarvo Added sortEmpty back in as an svg file and swapped ADataTable back to use it

brennarvo (approved):

PR #99, opened 2023-01-17
Rework table header for auto truncation

rwharrisjr:
This changes up the DOM structure to support auto truncation with text-overflow ellipsis. 

Note that this is potentially a breaking change with existing header css. Any existing overrides to this effect can hopefully be removed.

Use the `truncateHeaders` prop on `ADataTable` to set auto truncation. Basic example on demo is updated to use this prop

Screenshot 2023-01-17 at 1 47 01 PM
Screenshot 2023-01-17 at 1 47 10 PM

### DIFF
--- a/framework/components/ADataTable/ADataTable.scss
+++ b/framework/components/ADataTable/ADataTable.scss
@@ -177,29 +177,27 @@ $hidden-table-col-width: 45px;
   }
 
   &--truncate-headers {
-    thead .a-data-table__row {
-      .a-data-table__header {
-        max-width: 1px;
-        white-space: nowrap;
-        overflow: hidden;
-        text-overflow: ellipsis;
+    .a-data-table__header {
+      max-width: 1px;
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
 
-        &__sort {
-          &__button {
-            display: inline-flex;
-            overflow: hidden;
-          }
-        }
-
-        &__sort-wrap {
-          display: flex;
-        }
-
-        &__label {
-          text-overflow: ellipsis;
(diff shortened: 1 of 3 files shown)


framework/components/ADataTable/ADataTable.scss
```diff
@@ -184,4 +175,32 @@ $hidden-table-col-width: 45px;
...
     }
   }
+
+  &--truncate-headers {
+    thead .a-data-table__row {
+      .a-data-table__header {
+        max-width: 1px;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
```
brennarvo, line 185: Is there a situation where something other than the label class below would overflow and need ellipsis? I'm thinking we just need these within the `__label` class selector.

framework/components/ADataTable/ADataTable.scss
```diff
@@ -184,4 +175,32 @@ $hidden-table-col-width: 45px;
       }
     }
   }
+
+  &--truncate-headers {
+    thead .a-data-table__row {
```
brennarvo, line 180: I'm thinking we don't need to be this specific (leaving some room for an external class override).

```scss
&--truncate-headers {
  .a-data-table__header {
     max-width: 1px;
     // rest of styles
  }
}
```

rwharrisjr: Updated

framework/components/ADataTable/ADataTable.scss
rwharrisjr: Probably not. Cleaned up.

brennarvo (approved):

PR #103, opened 2023-01-18
More magnetic sync for AStepper

### LINKED ISSUE #100: Update AStepper to match Magnetic styling

AStepper appears to be using the wrong text color, icon size, font-size, font-weight, alignment, spacing, and inactive state styling. It also has a border after the steps and the Magnetic stepper does not. Hovering over a step should show `cursor: pointer` since it can be interactive in that clicking a step can take the user to that step. 

Example of magna vertical AStepper:

Screenshot 2023-01-17 at 9 23 38 PM

Example of Magnetic stepper:

Screenshot 2023-01-17 at 9 27 59 PM

Figma: https://www.figma.com/file/oVZWatImEIbl1c8sjdGxi0/branch/yU3qYkj6T2kfN4XYtUkEeD/%F0%9F%A7%B2--Magnetic-Design-Library?node-id=12114%3A49036&t=GKNxa1AtOblwz4it-0

A variation of AStepper that matches Magnetic so we can use it in our Incident Response feature: 
https://www.figma.com/file/ij1hK1tbZlee3RQQtM91Hb/Incidents?node-id=3248%3A49358&t=Xuo2mrmRIikhIJCb-0

rwharrisjr:
Resolves #100 

Note that the colors won't exactly line up since there is usage of two different color token sets in the design. Spacing, fonts and margins are also not 100%, but have been adjusted for a hopefully safe compromise between not breaking existing usage and getting somewhat closer to magnetic design.

### DIFF
--- a/framework/components/AStepper/AStep.js
+++ b/framework/components/AStepper/AStep.js
@@ -55,8 +55,11 @@ const AStep = forwardRef(
         setActiveStep && setActiveStep(stepNumber);
       },
       keyDownHandler = (e) => {
-        if (onClose && [keyCodes.enter, keyCodes.space].includes(e.keyCode)) {
-          onClick && onClick(e);
+        if (
+          setActiveStep &&
+          [keyCodes.enter, keyCodes.space].includes(e.keyCode)
+        ) {
+          onClick(e);
         }
       };
 
--- a/framework/components/AStepper/AStep.js
+++ b/framework/components/AStepper/AStep.js
@@ -15,6 +15,7 @@ const AStep = forwardRef(
       showIconOnVisited,
       stepNumber,
       children,
+      setActiveStep,
       ...rest
     },
     ref
@@ -32,6 +33,10 @@ const AStep = forwardRef(
       }
     }
 
+    if (!disabled && setActiveStep) {
+      className += ` a-step--cursor`;
+    }
+
     let style;
     if (
       React.Children.count(children) === 1 ||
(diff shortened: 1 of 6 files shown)


framework/components/AStepper/AStepper.scss
```diff
@@ -39,11 +39,15 @@
...
 
       &.a-step:after {
-        background: map-get($theme, "steps", "step-icon-border-color");
+        //  background: map-get($theme, "steps", "step-icon-border-color");
       }
       .a-step__label {
         color: map-deep-get($theme, "base", "color");
       }
+
+      .a-step__tilte {
```
sabrinamokerji, line 48: It looks like this might be a typo for `title`?

rwharrisjr: Good eye! Forgot to git add after I had updated that.

framework/components/AStepper/AStep.js
```diff
@@ -25,12 +25,18 @@ const AStep = forwardRef(
...
-      if (active) {
+      } else if (active) {
         className += ` a-step__active`;
+      } else {
+        className += ` a-step__default`;
       }
     }
 
+    let style;
+    if (children.length === 1 || (children[1] && !children[1].props.children)) {
```
brennarvo, line 36: I'm not sure if this is being applied correctly; right now in the UI, if I render just one child of `AStep`, it still looks like this with the margin:

Screenshot 2023-01-24 at 10 04 46 AM

I'm assuming that the reasoning for the `margin: 0` is to remove that (orange) margin so that it looks something like this (I hardcoded a zero margin into this example):

Screenshot 2023-01-24 at 10 05 28 AM

I think a better way of checking this is with the [React Top-Level API's `React.Children.count` method](https://reactjs.org/docs/react-api.html#reactchildrencount), e.g., `React.Children.count(children) === 1`

framework/components/AStepper/AStepper.scss
```diff
@@ -39,11 +39,15 @@
       }
 
       &.a-step:after {
-        background: map-get($theme, "steps", "step-icon-border-color");
+        //  background: map-get($theme, "steps", "step-icon-border-color");
```
brennarvo, line 42: Remove uncommented style

framework/components/AStepper/AStepper.scss
```diff
@@ -52,13 +56,34 @@
...
+      .a-step__hint {
+        color: map-deep-get($theme, "base", "step-default-color");
+      }
+    }
   }
   &--orientation-vertical {
     .a-step {
       &:after {
-        background-color: map-deep-get($theme, "steps", "step-line-color");
+        // background-color: map-deep-get($theme, "steps", "step-line-color");
```
brennarvo, line 74: Remove uncommented style

brennarvo (changes requested):

brennarvo (approved):
lgtm

rwharrisjr:
@sabrinamokerji This PR also adds a prop on `AStep` called `setActiveStep` - this will allow for clicking on steps to jump around in the stepper, and give control to specific use cases as needed.

PR #106, opened 2023-01-19
ASkeleton component

### LINKED ISSUE #71: Add Skeleton components for consistent loading state in apps
This was proposed during design of the Incident Manager table loading state. Skeleton loaders are fairly widely used now by apps. UX research shows it gives the user a perception of shorter loading times.

It allows us to put placeholder content where the real content would go, while a component is in a "loading state".

Magnetic Design proposal:
https://www.figma.com/file/FFnGj0SjnbC70fe0Ip1h5l/Component-Explorations?node-id=117%3A47661&t=GSsDfeHJw85y8ZQg-0

Implement common skeleton components for consistency across applications.

Loading spinners are the closest we have today.
Open source packages are available, these are a couple of easy to use React libraries, but we thought for consistency we'd want all apps to use a common version of something like this:
- [react-loading-skeleton](https://www.npmjs.com/package/react-loading-skeleton)
- [react-content-loader](https://www.npmjs.com/package/react-content-loader)

**Additional context**
From Magnetic:
image

Mockup of Incident Manager (Credit: Wendy Hsu)
image

rwharrisjr:
Resolves #71 

Screenshot 2023-01-19 at 1 37 16 PM

### DIFF
--- a/framework/components/ASkeleton/ASkeleton.js
+++ b/framework/components/ASkeleton/ASkeleton.js
@@ -45,7 +45,7 @@ const ASkeleton = forwardRef(
     );
 
     return (
-      <APanel className={className} {...rest}>
+      <APanel ref={ref} className={className} {...rest}>
         {header && <ASkeletonHeader />}
         {content}
       </APanel>
@@ -54,6 +54,10 @@ const ASkeleton = forwardRef(
 );
 
 ASkeleton.propTypes = {
+  /**
+   * String representing class names to be passed to component container
+   */
+  className: PropTypes.string,
   /**
    * Automatically add a header component
    */
--- a/framework/components/ASkeleton/ASkeleton.js
+++ b/framework/components/ASkeleton/ASkeleton.js
@@ -1,0 +1,77 @@
+import PropTypes from "prop-types";
+import React, {forwardRef} from "react";
+import {APanel} from "../APanel";
+import ASkeletonHeader from "./ASkeletonHeader";
+
+import "./ASkeleton.scss";
+
+const baseClass = "a-skeleton";
+
+const ASkeleton = forwardRef(
+  (
+    {
+      children,
+      className: propsClassName,
+      nested,
+      header,
+      horizontal,
+      animated,
+      ...rest
+    },
+    ref
+  ) => {
+    let className = baseClass;
+
+    if (propsClassName) {
+      className += ` ${propsClassName}`;
+    }
+
+    if (nested) {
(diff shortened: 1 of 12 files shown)


framework/components/ASkeleton/ASkeleton.js
```diff
@@ -0,0 +1,77 @@
...
+    return (
+      <APanel className={className} {...rest}>
+        {header && <ASkeletonHeader />}
+        {content}
+      </APanel>
+    );
+  }
+);
+
+ASkeleton.propTypes = {
```
brennarvo, line 56: Let's add the `className` to prop types here. I know it's probably intuitive that we allow custom classes to be passed in a UI library, but doesn't hurt to be pedantic in documentation.

framework/components/ASkeleton/ASkeleton.mdx
```diff
@@ -0,0 +1,102 @@
...
+    </ASkeleton>);
+    }`
+
+}
+
+/>
+
+## Skeleton Props
+
+The `ASkeleton` component inherits passed props.
```
brennarvo, line 80: Not related to this PR, but I've always wondered what this statement means in our documentation. In this case, I'm assuming it means that `ASkeleton` also inherits the props of `APanel` (since that is the root component it renders), but this line in the documentation doesn't necessarily make that too apparent.

framework/components/ASkeleton/ASkeleton.mdx
```diff
@@ -0,0 +1,102 @@
...
+        <ASkeletonText />
+      </ASkeleton>
+      <ASkeleton header nested style={{flex: "2 0 auto"}}>
+        <ASkeletonBlock />
+        <ASkeletonText />
+      </ASkeleton>
+    </ASkeleton>`}
+/>
+
+<Playground
```
brennarvo, line 59: This is a great example. I think we should also give it a heading, too, so that this specific example appears in the `ASkeleton` sidebar menu. Maybe something like "Horizontal Chunks"

brennarvo (changes requested):
Looks good, just a few of refinements.

framework/components/ASkeleton/ASkeleton.mdx
rwharrisjr: I think this is to indicate that it applies `{...rest}` to the parent element of the component. In this case it would make sense to specify it gets put on APanel.

brennarvo (approved):

PR #116, opened 2023-01-23
#111 - add size variants for AModal

### LINKED ISSUE #111: Add Magnetic modal sizes
Magnetic has 4 modal sizes: https://magnetic.cisco.com/0a43ab5cd/p/02a012-modal

However, the magna modal spans the full width of the screen:

Screenshot 2023-01-20 at 10 46 14 PM

Can we add the different size variations? Devs are making them all different sizes.

And if possible, can we also have a prop added that will center the modal in the middle of the screen instead of the top of the page?

rwharrisjr:
Resolves #111 

This PR adds a prop for each of the four sizes. Additionally, it adds `onClickOutside` so AModal can be a drop in replacement for ADialog

### DIFF
--- a/framework/components/AModal/AModal.js
+++ b/framework/components/AModal/AModal.js
@@ -28,6 +28,7 @@ const AModal = forwardRef(
       medium,
       large,
       xlarge,
+      onClickOutside,
       ...rest
     },
     ref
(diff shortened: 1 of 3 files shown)


brennarvo (approved):
Really nice examples for the new props. lgtm

PR #125, opened 2023-01-25
Add a cellLoader to auto use ASkeleton

rwharrisjr:
To make it a bit easier to add a loading state to `ADataTable`, this PR allows the use of a row data object `{cellLoading: true}`. This can be combined with row data objects that have data, and should allow for a full table loading state and infinite scroll situations.

Screenshot 2023-01-25 at 10 01 23 AM

### DIFF
--- a/framework/components/ADataTable/ADataTable.js
+++ b/framework/components/ADataTable/ADataTable.js
@@ -4,7 +4,7 @@ import React, {forwardRef, useMemo, useRef, useCallback, useState} from "react";
 import AInView from "../AInView";
 import AIcon from "../AIcon";
 import ASimpleTable from "../ASimpleTable";
-import ATooltip from "../ATooltip";
+import {ASkeleton, ASkeletonText} from "../ASkeleton";
 import "./ADataTable.scss";
 
 /**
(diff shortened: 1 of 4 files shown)


framework/components/ADataTable/ADataTable.js
```diff
@@ -325,16 +325,27 @@ const ADataTable = forwardRef(
                       </TableCell>
                     )}
                     {headers.map((y, j) => {
+                      console.log(y.cell, rowItem[y.key]);
```
brennarvo, line 328: Console log leak

framework/components/ASkeleton/ASkeleton.js
```diff
@@ -73,7 +76,11 @@ ASkeleton.propTypes = {
   /**
    * Show a shine keyframe animation to indicate loading
    */
-  animated: PropTypes.bool
+  animated: PropTypes.bool,
+  /**
+   * Use a <span> wrapper instead of `APanel`
+   */
+  noPanel: PropTypes.bool
```
brennarvo, line 83: Not that important (optional), but how do you feel about renaming this to `hidePanelBackdrop` or `removePanelBackdrop`? I know it's a bit long..but I wasn't 100% sure what`noPanel` referred to when first looking at the prop name.

brennarvo: Also, let's add an explicit example of this in the `ASkeleton` docs. I know it's already being showcased in the table loading example, but this will make it more apparent what's possible with the `ASkeleton` component.

brennarvo (changes requested):
Two changes I think we'll need, with another one just being a small naming convention nit.

brennarvo (approved):

PR #126, opened 2023-01-25
#124 - change button styles for first/last/next/prev in APagination

### LINKED ISSUE #124: Insufficient focus styles on pagination buttons
**Describe the bug**
Makes it hard to navigate by keyboard due to unclear focus. 
image

Similarly on hover, there's a hand cursor change, but it would be better to have also some kind of visual change of the element.

**To Reproduce**
Hover of focus the arrow buttons.

**Expected behavior**
Clearer hover and focus indication. I'm not sure what the exact gudelines for hover/focus are, but I would guess it could look the same/similar to input states:
image
image
image

Or it could use tertiaty button:
image

Real button would also improve this variant. The arrows look way to tiny and close to each other for magnetic visual style:
image

Or it could be somehow consistent with numbers in https://www.figma.com/file/oVZWatImEIbl1c8sjdGxi0/branch/yU3qYkj6T2kfN4XYtUkEeD/%F0%9F%A7%B2--Magnetic-Design-Library?node-id=5247%3A39676&t=UptKVbwAZp2EycjO-0, but I would use different visual for active page number and arrow button focus/hover.

 - Device: Desktop
 - OS:  iOS, Ubuntu
 - Browser Chrome
 - Version latest

**Additional context**

rwharrisjr:
Resolves #124

### DIFF
--- a/framework/components/APagination/APagination.js
+++ b/framework/components/APagination/APagination.js
@@ -118,7 +118,7 @@ const APagination = forwardRef(
               <AButton
                 className="a-pagination__first"
                 disabled={page === 1}
-                tertiaryAlt
+                tertiary
                 icon
                 onClick={() => onPageChange(1)}
                 aria-label="First"
@@ -129,7 +129,7 @@ const APagination = forwardRef(
             <AButton
               className="a-pagination__previous"
               disabled={page === 1}
-              tertiaryAlt
+              tertiary
               icon
               onClick={() => onPageChange(page - 1)}
               aria-label="Previous"
@@ -180,7 +180,7 @@ const APagination = forwardRef(
             <AButton
               className="a-pagination__next"
               disabled={page === pages}
-              tertiaryAlt
+              tertiary
               icon
               onClick={() => onPageChange(page + 1)}
               aria-label="Next"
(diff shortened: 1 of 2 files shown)


brennarvo (approved):
This is a good change. [Magnetic Figma mock-ups](https://www.figma.com/file/oVZWatImEIbl1c8sjdGxi0/%F0%9F%A7%B2--Magnetic-Design-Library?node-id=5247%3A39676&t=q1R41To9lXCLFifx-0) don't currently have anything mentioning hover states on the arrow buttons, so I think this is our best direction forward. It definitely improves the accessibility of interacting with the pagination component.

mvelkCisco:
Looking into this again, keeping the `tertiary-alt` might be actually better, as long as the icon changes color on hover/focus.
It does on hover here: https://magna-react.vercel.app/components/button#variants
image
image
so I assume there's some kind of override here (especially when `tertiary` should have a blue icon).

However `tertiary-alt` do not change the color right now on focus. I was gonna add an issue for that.

It's definitely better that the icons are bigger. But if you want to add a background, please add a margin between buttons too. And it would look better if they were sqare.
image

Issue #128, opened 2023-01-25
Allow the user to set dimensions/size for the `AEmptyState` component


vdineva:
I'm adding the `<AEmptyState />` component and it seems disproportionately big to rest of the elements on the page: 
![Screenshot 2023-01-25 at 13 58 48](https://user-images.githubusercontent.com/791418/214658419-326e88ee-4ca6-4992-9a93-a1a52ce76b52.png)

I tried passing in inline styles but with no success so far. It would be great if the width/height of this component can be easily adjusted

rwharrisjr:
Resolved in #129

PR #129, opened 2023-01-27
#128 - add size variants for AEmptyState

rwharrisjr:
Screenshot 2023-01-27 at 1 05 32 PM
Screenshot 2023-01-27 at 1 05 53 PM

### DIFF
--- a/framework/components/AEmptyState/AEmptyState.js
+++ b/framework/components/AEmptyState/AEmptyState.js
@@ -15,17 +15,21 @@ const AEmptyState = forwardRef(
       label,
       message,
       children,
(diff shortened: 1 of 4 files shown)


framework/components/AEmptyState/AEmptyState.js
```diff
@@ -46,6 +50,18 @@ const AEmptyState = forwardRef(
       icon = "info";
     }
 
+    if (small) {
```
androche, line 53: switch case might look a little cleaner, but nbd if there's a standard already

androche (approved):

Issue #84, opened 2023-01-09
Replace atomic icons with Phosphor icons where applicable (list)


derrick-snider:
### Description
Increase consistency and alignment with Magnetic by updating all possible icons using the Phosphor icons. In the future it would be great to utilize the Phosphor icon font to utilize the different weights and styles (fill/2 tone...). Referring to this gallery list https://magna-react.vercel.app/components/icon and https://phosphoricons.com/ .

### Checklist
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`
- [ ] current "iconname" replace with `phosphor`

- [ ] current "iconname" replace with `phosphor`

### Images
gallery view example
<img width="1079" alt="image" src="https://user-images.githubusercontent.com/429803/211349204-2f3e9a55-e845-4f3d-9238-d580cd0568d4.png">

phosphor name example
<img width="654" alt="image" src="https://user-images.githubusercontent.com/429803/211350039-b0e23cf8-e01c-4a4b-a21f-77af3ec745c2.png">

### CC

atjackson49:
- [X] replace "add" with plus
- [ ]  replace "add-to-list" with file-plus
- [ ]  replace "alarm" with alarm
- [X]  replace "alert" with bell
- [ ]  replace "align-center" with text-align-center
- [ ]  replace "align-left" with text-align-left
- [ ]  replace "align-right" with text-align-right
- [ ]  replace "allow" with arrow-circle-right
- [ ]  replace "allow-list" with list-checks
- [ ]  keep "api-cloud" for now
- [ ]  replace "append" with arrow-elbow-right-up
[...]

PR #141, opened 2023-01-30
Reconcile phosphor icons

rwharrisjr:
This is a pass at bringing in icons and replacing what was previously provided by Atomic with the Phosphor https://phosphoricons.com/ icon set. 

Icon names not available in Phosphor that were previously available in Atomic will still fallback to the original Atomic icon.

Note that this is a breaking change, as icon names for `AIcon` now map directly to Phosphor names. Any usage of `iconSet="magna"` should be removed and the icon name converted to hyphenated rather than camel case.


scripts/copyPhosphorIcons.mjs
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Run `npm run copyPhosphorIcons`
+ */
+
+const phosphorPath = `${os.homedir()}/Downloads/phosphor-icons/SVGs/Regular`;
+const outputBase = "./svg";
+
+const {iconNameMap} = icons;
+
+const generatedListForPhosphor = Object.keys(iconNameMap).map((key) => {
```
brennarvo, line 23: This might be a use-case for `Object.entries`:

```js
const generatedListForPhosphor = Object.entries(iconNameMap).map(
  ([atomicKey, phosphorKey]) => phosphorKey || atomicKey
);
```

Also, looking at how `generatedListForPhosphor` is used, could this just be a `filter`? With `map`, it will also include atomic icon names that don't resolve to any phosphor name, which means the ```path.resolve(`${phosphorPath}/${key}.svg`)``` on line 36 will fail (albeit it is being caught — I'm just not sure if that was the intent).

rwharrisjr: The intent was for it to fail with a warning - it gives us a way to know at the script run time which icons are falling back to atomic. In the future we may check for icons that match up and replace if needed.

 Updated code to use suggested `Object.entries`

framework/components/AIcon/AIcon.js
```diff
@@ -19,7 +15,7 @@ const AIcon = forwardRef(
       left,
       right,
       size,
-      iconSet = ICON_SETS.ATOMIC,
+      iconSet,
```
brennarvo, line 18: Wen can remove this prop now

framework/components/AIcon/AIcon.js
```diff
@@ -55,16 +73,21 @@ const AIcon = forwardRef(
     if (iconDef) {
       const {props: iconProps, xml} = iconDef;
 
+      if (ignoreStrokeReplace.includes(children)) {
+        console.log(children);
```
brennarvo, line 77: If this isn't intentional, we can remove

framework/components/AIcon/AIcon.js
```diff
@@ -55,16 +73,21 @@ const AIcon = forwardRef(
...
+        style.stroke = "none";
+      }
+
       return (
-        <svg {...iconProps} {...componentProps}>
+        <svg {...iconProps} {...componentProps} style={style}>
```
brennarvo, line 82: Adding `style` here will override the one defined on the `componentProps` object, which is used for a custom `size` prop (see lines 58-60).

We can just add the `style.stroke` assignment to `componentProps.style.stroke` instead.

scripts/copyPhosphorIcons.mjs
```diff
@@ -15,7 +15,7 @@ import icons from "../framework/components/AIcon/atomicMap.js";
  * Run `npm run copyPhosphorIcons`
  */
 
-const phosphorPath = `${os.homedir()}/Downloads/phosphor-icons/SVGs/Fill`;
+const phosphorPath = `${os.homedir()}/Downloads/phosphor-icons/SVGs/Regular`;
```
brennarvo, line 18: The way I interpreted the requirements in #84  was that we need to support both `fill` and `stroke`.

brennarvo (changes requested):

brennarvo (approved):
LGTM 👍 Next beta release, we'll want to make sure to mention how the icon names need to be converted to kebab-case.

PR #142, opened 2023-01-30
Null check AModal portal node

rwharrisjr:


brennarvo (approved):
Thanks for looking into this

PR #147, opened 2023-02-03
#122 - add focus/hover state to data table sort headers

### LINKED ISSUE #122: Missing hover and focus styles table header cells with sort
**Describe the bug**
Table header cells that support sorting have no hover of focus style. Especially due to missing focus styling it's not clear where the focus is when navigating by keyboard.

**To Reproduce**
Hover of focus a sortable column by keyboard.

**Expected behavior**
There should be some visual distinction for focus and hover. 

I'm guessing it should be done by the blue line above the cell as indicated in variants here https://www.figma.com/file/oVZWatImEIbl1c8sjdGxi0/branch/yU3qYkj6T2kfN4XYtUkEeD/%F0%9F%A7%B2--Magnetic-Design-Library?node-id=5259%3A38404&t=UptKVbwAZp2EycjO-0. But I'm not sure, needs confirmation.

In that case it would be great if the whole cell was clickable, not just the text (or show the line only when hovering the text).

**Screenshots**
image

 - Device: Desktop
 - OS: macOS
 - Browser Chrome
 - Version latest

rwharrisjr:
Resolves #122

### DIFF
--- a/framework/components/ADataTable/ADataTable.scss
+++ b/framework/components/ADataTable/ADataTable.scss
@@ -16,7 +16,7 @@ $hidden-table-col-width: 45px;
       cursor: pointer;
       &:focus,
       &:hover {
-        border-top: 2px solid map-deep-get($theme, "base", "color--active") !important;
+        border-top: 2px solid map-deep-get($theme, "table", "header-focus") !important;
       }
     }
   }
--- a/framework/components/ADataTable/ADataTable.scss
+++ b/framework/components/ADataTable/ADataTable.scss
@@ -9,6 +9,16 @@ $hidden-table-col-width: 45px;
         color: map-deep-get($theme, "base", "color");
       }
     }
+
+    border-top: 2px solid transparent;
+    border-top-color: transparent;
+    &--sortable {
+      cursor: pointer;
+      &:focus,
+      &:hover {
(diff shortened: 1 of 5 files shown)


framework/components/ADataTable/ADataTable.scss
```diff
@@ -9,6 +9,16 @@ $hidden-table-col-width: 45px;
...
       }
     }
+
+    border-top: 2px solid transparent;
+    border-top-color: transparent;
+    &--sortable {
+      cursor: pointer;
+      &:focus,
+      &:hover {
+        border-top: 2px solid map-deep-get($theme, "base", "color--active") !important;
```
brennarvo, line 19: The figma for the table header borders aren't resolving to any real hex values..but it looks like it should be closer to something like `#306ACF`, i.e., `$mds-interactive-10` in our palette.

brennarvo (changes requested):

rwharrisjr:
Screenshot 2023-02-14 at 9 16 16 AM


Looking lower in the component the blue there, which is of similar function I think, comes out to `interactive-8`, so I updated to that. For dark theme the color doesn't exist in our palette, but `interactive-8` is close enough for now.

brennarvo (approved):

PR #153, opened 2023-02-07
#152 - fix border radius for first and last menu item

### LINKED ISSUE #152: Slight whitespace on border-radius gap
Slight whitespace on menu items on hover
Screenshot 2023-02-06 at 10 03 59 AM

rwharrisjr:
Resolves #152 

Note that this breaks the example with custom styling for border radius on the menu, but does fix the default rounded border case.

See "More Menu" button on the Storybook

### DIFF
--- a/framework/components/AMenu/AMenu.scss
+++ b/framework/components/AMenu/AMenu.scss
@@ -1,5 +1,7 @@
 @import "../../styles";
 
+$menu-item-first-last-border-radius: 4px;
+
 /*
 When using AMenu that passes AMenuBase to AList, force override the specificity
 since the css may be imported in different orders depending on product build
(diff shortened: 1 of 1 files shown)


brennarvo (commented):
Another option is to add `overflow: hidden` to the menu container and remove the `border-radius` entirely from the menu items so that their corners are covered by the parent menu's container. My only worry is that maybe there are some instances out there right now where content *is* actually overflowing the menu on purpose?

rwharrisjr:
I can't think of a use case for intentionally overflowing the menu. I think adjusting the border radius guarantees non breaking, even though I like your solution better.

androche:
I like pictures :)

rwharrisjr:
Before

Screenshot 2023-02-09 at 1 55 50 PM


After
Screenshot 2023-02-09 at 1 55 22 PM

androche (approved):

PR #154, opened 2023-02-07
Remove AHeader component

rwharrisjr:
Should not be used. Replacement is the CDS header.


brennarvo (commented):
Can we leave the MDX file and add [a link to the CDS header](https://ciscodesignsystems.github.io/magnetic-common-design-system/?path=/story/react-components-header-documentation--implementation) so that people aren't confused why it went away?

brennarvo (approved):
Looks good. I also made a PR to remove the component from the eslint config: https://github.com/cisco-sbg-ui/eslint-config-magna-react/pull/7

PR #159, opened 2023-02-08
refactor: add medium button styles to sass mixin

brennarvo:
I think a mixin for the button size makes sense so that we can reuse it where needed, but this could also seem a bit excessive. I just worry about the `AButtonGroup` and `AButton` having different declarations for the same thing.


rwharrisjr (approved):
nice!

PR #162, opened 2023-02-09
fix: listen for mousedown events in useOutsideClick

### LINKED ISSUE #143: `usePopupQuickExit` closes `ADrawer` mistakenly when using `AMenu` to Open
**Describe the bug**
`ADrawer` does not open when using an option in an `AMenu` to open the drawer mixed with `usePopupQuickExit` usage.

The `usePopupQuickExit`'s `onExit` callback is getting triggered to immediately close the drawer when trying to open. See the steps below to reproduce on the magna-react docs page.

Ultimately its `useOutsideClick` causing `onExit` to be called.

**To Reproduce**
1. Go to https://magna-react.vercel.app/components/drawer
2. Paste the following code snip into the first code editor block in the Drawer docs page:

```jsx
() => {
  const button3Ref = useRef(null);
  const [open3, setOpen3] = useState(false);

  const nonModalContentRef = useRef();
  const [nonModalOpen, setNonModalOpen] = useState(false);
  usePopupQuickExit({
    popupRef: nonModalContentRef,
    isEnabled: nonModalOpen,
    onExit: () => { setNonModalOpen(false); console.log('CLOSING DRAWER'); }
  });

  return (
    <>
      <p>Non-Modal Drawer Opened with a Menu Option</p>
      <AButton ref={button3Ref} onClick={() => setOpen3(true)}>Open Menu</AButton>
      <AMenu
        anchorRef={button3Ref}
        open={open3}
        placement="bottom-left"
        onClose={() => setOpen3(false)}
        style={{borderRadius: 0}}>
        <AListItem onClick={() => setNonModalOpen(true)}>Open Drawer</AListItem>
      </AMenu>
      <ADrawer
        ref={nonModalContentRef}
        asModal={false}
        isOpen={nonModalOpen}
        slideIn="right"
      >
          Hello World
      </ADrawer>
    </>
  );
};
```
3. Click the Button to Open the Menu
4. In the Menu use the Option to open the Drawer
5. Notice that the Drawer does not open
6. Check console & see log showing `onExit` was called when trying to open the Drawer

**Expected behavior**
The Drawer should be able to open and stay open with a dropdown menu. The drawer's `usePopupQuickExit` should not depend on the element used to open the drawer to be in the DOM.

**Screenshots**

https://user-images.githubusercontent.com/95231401/215842468-38d2344a-228f-4f38-b228-0f679bab6486.mov

 - Device: Desktop
 - OS: `MacOS`
 - Browser `Chrome, Safari, FF`
 - Version: `1.0.0-beta.13`

**Additional context**

This seems to happen when the element that is clicked on to open the drawer is removed from the DOM.

brennarvo:
Closes #143 

`useOutsideClick` used the `click` event listener, which is fired *after* the `mousedown` listener, meaning that there were possibilities for race conditions as it related to components unmounting within the `useOutsideClick`'s container element.

This PR updates `useOutsideClick` to listen to `mousedown` events, meaning that we will be alerted about outside clicks to the container element *before* the user finishes their click. 

[MDN's note about the `mousedown` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousedown_event)

> Note: This differs from the [click](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) event in that click is fired after a full click action occurs; that is, the mouse button is pressed and released while the pointer remains inside the same element. mousedown is fired the moment the button is initially pressed.

### DIFF
--- a/framework/hooks/useOutsideClick.js
+++ b/framework/hooks/useOutsideClick.js
@@ -18,10 +18,10 @@ const useOutsideClick = (options) => {
     if (!isEnabled) {
       return;
     }
-    window.addEventListener("click", detectOutside);
+    window.addEventListener("mousedown", detectOutside);
 
     return () => {
-      window.removeEventListener("click", detectOutside);
+      window.removeEventListener("mousedown", detectOutside);
     };
   }, [isEnabled, onClick, detectOutside]);
 };


rwharrisjr (approved):
Tested with code provided in issue, seems to work

PR #165, opened 2023-02-13
#163 - make single level breadcrum arrow clickable

### LINKED ISSUE #163: Back arrow in single level breadcrumb is not clickable
In [single level breadcrumb](https://magna-react.vercel.app/components/breadcrumb#usage-single-level), the left arrow is not clickable, doing nothing. Only the provided component/link is clickable.

For me the expected and common behaviour on the web is when there is a single breadcrumb/back-button, the whole element is clickable including any icons or idiomatic graphical symbols.

rwharrisjr:
Resolves #163

### DIFF
--- a/framework/components/ABreadcrumb/ABreadcrumb.js
+++ b/framework/components/ABreadcrumb/ABreadcrumb.js
@@ -6,7 +6,14 @@ import "./ABreadcrumb.scss";
 
 const ABreadcrumb = forwardRef(
   (
-    {className: propsClassName, item, items = [], magnetic = false, ...rest},
+    {
+      className: propsClassName,
+      item,
+      items = [],
+      magnetic = false,
+      onClick,
+      ...rest
+    },
     ref
   ) => {
     let className = `a-breadcrumb`,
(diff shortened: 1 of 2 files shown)


framework/components/ABreadcrumb/ABreadcrumb.js
```diff
@@ -6,7 +6,14 @@ import "./ABreadcrumb.scss";
 
 const ABreadcrumb = forwardRef(
   (
-    {className: propsClassName, item, items = [], magnetic = false, ...rest},
+    {
+      className: propsClassName,
+      item,
+      items = [],
+      magnetic = false,
+      onClick,
```
brennarvo, line 14: I don't think this is being used? Looks like `onClick` is supposed to be added to the `contentProps` for the objects in `items`.

framework/components/ABreadcrumb/ABreadcrumb.js
```diff
@@ -71,6 +97,9 @@ ABreadcrumb.propTypes = {
    * Adjust for magnetic sizing
    */
   magnetic: PropTypes.bool
+  /**
```
brennarvo, line 100: Probably where `onClick` was supposed to go (if still needed).

framework/components/ABreadcrumb/ABreadcrumb.js
rwharrisjr: Good catch, realized it was on the item after I started and missed the cleanup since linter didn't pick it up.

rwharrisjr:
Updated the mdx doc

brennarvo (approved):

PR #167, opened 2023-02-13
#166 - move validation state css below readOnly

### LINKED ISSUE #166: Text input with readOnly flag does not reflect border color change caused by validationState
**Describe the bug**
I'm working on an input component with an attached date picker:
image
The input is set to be `readOnly` to simplify the validation. 

When showing a `hint` with `validationState="danger"` the input does not get a red border by default:
image

It does turn red when hovered or focused:
image

This is inconsistent with other inputs:
image

**Expected behavior**
Border should be red for `readOnly` inputs when not hovered/focused. Even though the value can't be changed directly, it should still support the error state properly.

 - Device: Desktop
 - OS: macOS
 - Browser: Chrome
 - Version: Latest

rwharrisjr:
Resolves #166

### DIFF
--- a/framework/components/AInputBase/AInputBase.scss
+++ b/framework/components/AInputBase/AInputBase.scss
@@ -30,6 +30,17 @@ $input-transition: border-color $transition-duration--extra-fast
     }
   }
 
+  &--readonly {
+    .a-input-base__surface {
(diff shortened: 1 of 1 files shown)


brennarvo (approved):
The "cascade" in its glory

PR #173, opened 2023-02-15
#168 - add medium variant for ASelect

### LINKED ISSUE #168: ASelect medium variant (magnetic size) does not make options taller
**Describe the bug**
What `ASelect` with `medium` prop looks like:
image
Options are too narrow.

How it should look like:
image
Reference: https://magnetic.cisco.com/0a43ab5cd/p/42131c-select/t/74d621

Also icons and paddings don't match 100% but the option size is the most obvious.

**Environment:**
 - Device: Desktop
 - OS: macOS
 - Browser: Chrome
 - Version: Latest

rwharrisjr:
Resolves #168

### DIFF
--- a/framework/components/AList/AList.scss
+++ b/framework/components/AList/AList.scss
@@ -127,7 +127,11 @@
 
   &--medium {
     .a-list-item {
-      height: 36px;
+      height: $input-height-medium;
+
+      &::after {
+        top: 9px !important;
+      }
     }
   }
 }
--- a/framework/components/AList/AList.scss
+++ b/framework/components/AList/AList.scss
@@ -124,4 +124,10 @@
   a.a-list-item--two-line {
     text-decoration: none;
   }
+
+  &--medium {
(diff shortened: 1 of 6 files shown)


framework/components/AList/AList.scss
```diff
@@ -124,4 +124,10 @@
   a.a-list-item--two-line {
     text-decoration: none;
   }
+
+  &--medium {
+    .a-list-item {
+      height: 36px;
```
brennarvo, line 130: Is there a reason it's `36px` instead of `34px` like it is [inside `AInputBase`](https://github.com/cisco-sbg-ui/magna-react/blob/d5666b5c77054becd6da1abfd4ac5701a3a617ee/framework/components/AInputBase/AInputBase.scss#L171)?

Maybe also a good time to have a sass variable for the `medium` size that we can use across all the components.

brennarvo (approved):

PR #186, opened 2023-02-21
#183 - custom template options for ASelect

### LINKED ISSUE #183: Add support for a custom display template in ASelect input

Currently `ASelect` seems to be only able to render either a placeholder, or a value from `items`. 
Reference: https://github.com/cisco-sbg-ui/magna-react/blob/canary/framework/components/ASelect/ASelect.js#L333

I'm implementing a custom select using `ASelect` that supports infinite loading of data on scroll. It basically makes an API call for the next page when the scrolling reaches a skeleton appended to the end of `items`.

Video preview: https://user-images.githubusercontent.com/51229231/219660607-4bfce65a-fa62-44b8-8ae4-aa26edf3b757.mov

Code reference (`AsyncSelect`): https://github.com/advthreat/intelligence/pull/67/files#diff-6f78b0603173c76b71b4403406099fe4e014e2f1f0ccb917d6381a09af49da18

The issue is that user can select a value that's not on the first page of the loaded data. Next time the page loads, the `items` don't contain the selected value, so it won't get displayed as selected (just the placeholder renders).

A prop on `ASelect` that would allow setting a custom templating function similar to `itemTemplate`, but for the content of the select input. This would allow not only setting a custom display value (even not included in `items`), but also rich formatting like this:
image

Or at least a custom string value would be nice.

I figured out I can include the selected items that were not loaded to the `items` and then hide it using `css` in the `itemTemplate`.

See `hiddenSelectedItemsNotInLoadedItems`: https://github.com/advthreat/intelligence/pull/67/files#diff-6f78b0603173c76b71b4403406099fe4e014e2f1f0ccb917d6381a09af49da18R126

Template hiding: https://github.com/advthreat/intelligence/pull/67/files#diff-6f78b0603173c76b71b4403406099fe4e014e2f1f0ccb917d6381a09af49da18R78 

It seems to work, but it's super hacky.

rwharrisjr:
Resolves #183 

Adds

Screenshot 2023-02-21 at 9 36 45 AM

### DIFF
--- a/framework/components/ASelect/ASelect.js
+++ b/framework/components/ASelect/ASelect.js
@@ -326,7 +326,7 @@ const ASelect = forwardRef(
         selectionContent = (
           <MenuItemComponent
             key={`a-select__menu-item_selectedIndex`}
-            item={selectedDisplayItem || selectedItem}
+            item={itemContent}
           />
         );
       } else {
--- a/framework/components/ASelect/ASelect.js
+++ b/framework/components/ASelect/ASelect.js
@@ -43,6 +43,9 @@ const ASelect = forwardRef(
       validateOnBlur,
       validationState,
       medium,
+      useTemplateForSelectedItem = false,
+      selectdDisplayTemplate,
+      selectedDisplayItem,
       ...rest
     },
     ref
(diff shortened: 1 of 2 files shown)


mvelkCisco:
Thanks @rwharrisjr. I'll check it out. On the first sight though, it looks like there's a typo in `selectdDisplayTemplate`.

mvelkCisco:
Sorry for the delay, I'm just getting back to this. The proposed solution look very flexible, thank you!

brennarvo (changes requested):
This is a nice enhancement. Only thing might be the prop name confusion.

`selectedDisplayTemplate` and `selectedDisplayItem` seem similar to me. If I'm understanding it correctly, `selectedDisplayItem` should match the structure of an object/string in the `items` prop, whereas `selectedDisplayTemplate` is similar to `itemTemplate` in that it's a way to pass a custom wrapper component to what gets rendered.

Maybe we rename `selectedDisplayItem` to just `selectedItem` to make it clear that is the item object/string you can pass in.

Also just to note, with the current logic, the only way you can get a `selectedItem` (I'm using the name suggestion mentioned above) to appear is if you are also using `selectedDisplayTemplate`, but it might be handy to allow a `selectedItem` no matter what.

---

Come to think of it, I see why it's named `selectedDisplayItem`; it isn't actually the item that is stored in `ASelect` state. It's the item (to use the word itself) displayed in the UI.

I'm keeping the text above this section as a reminder to myself if I ever come back..but I'm also not sure if this is is a sign it might be confusing in general.

Either way, one typo change that should get fixed, then we're good 👍

PR #187, opened 2023-02-21
#184 - focus menu on open ASelect

### LINKED ISSUE #184: ASelects (sometimes) changes selection on keyboard up/down presses instead of just focusing the next/previous option
**Describe the bug**

This is a 2 in 1 bug (because of the "sometimes" part).

When an item is selected in `ASelect`, opening the select and hitting the up/down keys changes the selected item immediately. I think this is the intended behavior, judging based on the code: https://github.com/cisco-sbg-ui/magna-react/blob/canary/framework/components/ASelect/ASelect.js#L262.

However I don't think that's how a native select behaves. The up/down keys should highlight/focus the option, then user should confirm selection with Enter/Spacebar. See https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select.

The most confusing part is that it behaves inconsistently for me over time, as I interact with it. This part I don't understand from the code. Try:

1. fresh load https://magna-react.vercel.app/components/select and don't touch anything
2. click the "Fruit" in the select in the first example
3. hit up and then down keys -> observe selection changing right away 
4. click the page background to close the select
5. click the same select again to open it
6. hit up and then down keys -> selection not changing anymore, just highlight and focus (which I would expect in 3. as well)

Here's a video: https://user-images.githubusercontent.com/51229231/219678163-bcf3f29b-869a-428e-bc4a-36d8a5e094ae.mov

This also causes issues for my `AsyncSelect` mentioned in #183. The selected item might not be even loaded. By hitting the up/down keys, the selection is changed immediately, without any possibility for going back (originally selected item is not on the screen). 

To sum up: The selection should never change just by hitting up/down keys. Just the focus. Enter/Spacebar confirmation should be required to confirm the selection. And it should behave like this consistently.

 - Device: Desktop
 - OS: macOS
 - Browser: Chrome
 - Version: Latest

rwharrisjr:
Resolves #184

### DIFF
--- a/framework/components/ASelect/ASelect.js
+++ b/framework/components/ASelect/ASelect.js
@@ -265,7 +265,6 @@ const ASelect = forwardRef(
           } else if (e.keyCode === keyCodes.down) {
             e.preventDefault();
             const newItem = getNextItem(getSelectedIndex());
-            console.log(newItem);
             newItem && selectItem(newItem);
           }
         };
--- a/framework/components/ASelect/ASelect.js
+++ b/framework/components/ASelect/ASelect.js
@@ -69,10 +69,12 @@ const ASelect = forwardRef(
     }, [validationState]);
 
     useEffect(() => {
-      if (isOpen && selectedItemRef.current) {
-        selectedItemRef.current.focus();
+      if (isOpen && menuRef.current) {
+        setTimeout(() => {
+          menuRef.current.focus();
+        }, 0);
       }
-    }, [isOpen]);
+    }, [isOpen, menuRef]);
 
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


framework/components/ASelect/ASelect.js
```diff
@@ -263,6 +265,7 @@ const ASelect = forwardRef(
           } else if (e.keyCode === keyCodes.down) {
             e.preventDefault();
             const newItem = getNextItem(getSelectedIndex());
+            console.log(newItem);
```
brennarvo, line 268: Console.log here

brennarvo (changes requested):
One console.log to remove. But great catch and fix.

PR #205, opened 2023-02-28
Add required asterisk and infoTooltip to AFieldBase

### LINKED ISSUE #200: Support * and tooltip in label for required fields
Screenshot 2023-02-27 at 11 14 20 AM

Looking at: https://www.figma.com/file/oVZWatImEIbl1c8sjdGxi0/%F0%9F%A7%B2--Magnetic-Design-Library?node-id=5193%3A37461&t=Ej2edsCyhx1zK7nd-11

For the sake of consistency, we should support the * when the input is required. I'm not totally sure about the use of the "(optional)" marker. From designs I've seen, it is very rare so I would think maybe this should not be implemented at this point as it is implied if there is no * 

It would also be good to include support for a tooltip if the input includes the prop

rwharrisjr:
Resolves #200 

- Passes `required` to `AFieldBase` to automatically add red asterisk (*) next to the input label
- Adds new prop `infoTooltip` of type string, when passed will add an `AIcon` `information` icon next to the label with a tooltip


framework/components/AFieldBase/AFieldBase.js
```diff
@@ -71,6 +102,14 @@ AFieldBase.propTypes = {
...
    */
   labelId: PropTypes.string,
+  /**
+   * Adds required asterisk next to label
+   */
+  required: PropTypes.bool,
+  /**
+   * Adds an information tooltip next to the label
+   */
+  infoTooltip: PropTypes.string,
```
brennarvo, line 112: Unfortunately, these won't get added into the `ATextInput` docs: https://magna-react-git-issue-200.securex-preview.app/components/text-input#text-input-props

One pattern I've seen in the Incident Manager Repo to help with this is do import the inherited props and spread them onto the component that inherits those props:

```jsx
// ATextInput.js
import AFieldBase from '../AFieldBase';

const ATextInput = (props) => {
  return <AFieldBase {...props} />
}

const {...AFieldBaseProps} = AFieldBase.propTypes;

ATextInput.propTypes = {
  // All of `AFieldBase` props
  ...AFieldBaseProps,

  // Stuff specific to `ATextInput`
};
```

scripts/copyPhosphorIcons.mjs
```diff
@@ -31,15 +31,16 @@ const generatedListForPhosphor = Object.entries(iconNameMap).map(
 
 const copyIconFiles = () => {
   generatedListForPhosphor.forEach((key) => {
+    let outputName = key;
     if (key === "info") {
-      return;
+      outputName = "information";
```
brennarvo, line 36: Just want to confirm; this is becuase we've preserved the `info` name in our icon gallery for the blue info icon, right?

info

If that's the case, I'm wondering if it would make sense to prepend those colorful status icons with `status-*`. Maybe not for this PR..but just a thought.

brennarvo (changes requested):
One quick enhancement for the `ATextInput` prop types

scripts/copyPhosphorIcons.mjs
rwharrisjr: Right. Probably would be good to prepend with `status-`, that would cut down on future confusion

brennarvo (approved):

PR #225, opened 2023-03-02
fix AInputBase error and warning state focus effect

rrubas:


rrubas:
Screenshot 2023-03-02 at 12 54 07
Screenshot 2023-03-02 at 12 53 43

mvelkCisco:
It would be better to have a support for both light and dark theme via variables. Converting to draft.

mvelkCisco:
After further inspection, there are inconsistencies between magna-react and magnetic, but also inside magnetic figma as well. I've addd https://github.com/cisco-sbg-ui/magna-react/issues/229 for further investigation and would like to follow up with magnetic team. 

The color inconsistency is not a big issue visually but it's causing a lot of confusion for us about which colors to use. For now I would suggest to merge this PR to remove the glow effect at least.

rwharrisjr (approved):

PR #228, opened 2023-03-02
Use theme icon colors rather than currentColor

rwharrisjr:
@mvelkCisco pointed out that use of `currentColor` as the default doesn't line up with the magnetic designs. This was a carryover from the original atomic-react repo. 

This PR changes to use the theme mapped colors. Preview deployment for validation https://magna-react-git-icon-colors.securex-preview.app/components/icon

-----------

Also adding "kebab helper" so that most uses of camelCase icon names should continue to work 

Screenshot 2023-03-02 at 4 43 27 PM


mvelkCisco:
You are right that going forward this change will make us avoid more issues than it could introduce now. However id does break icon button:
https://magna-react-git-icon-colors.securex-preview.app/components/button#install
image

Including hover states. Those might need som kind of override.

mvelkCisco:
Tags before:
image

After:
image

mvelkCisco:
I quickly scanned the rest of the framework and I think fixing the `AButton` including the hover/focus colors, and the tags, should address all the regressions.

It might still break apps that relied on the original behavior, but I would still do the change.

rwharrisjr:
Fixed button and tag specific icons. 

Also added negative status fill for `critical-stop`. This was not colored correctly in the original atomic-react.

mvelkCisco:
This icon looks broken after the latest change https://magna-react-git-icon-colors.securex-preview.app/components/popover
image

rwharrisjr:
Good catch, had lost specificity on stroke unset for the legacy atomic icons.

brennarvo (approved):

PR #230, opened 2023-03-02
fix ABreadcrumb style

rrubas:
Breadcrumb now folow magnetic typography and hover state

### DIFF
--- a/framework/components/ABreadcrumb/ABreadcrumb.js
+++ b/framework/components/ABreadcrumb/ABreadcrumb.js
@@ -5,18 +5,11 @@ import AIcon from "../AIcon";
 import "./ABreadcrumb.scss";
 
 const ABreadcrumb = forwardRef(
-  (
-    {className: propsClassName, item, items = [], magnetic = false, ...rest},
-    ref
-  ) => {
+  ({className: propsClassName, item, items = [], ...rest}, ref) => {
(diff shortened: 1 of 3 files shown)


rrubas:
Screenshot 2023-03-02 at 15 43 25
Screenshot 2023-03-02 at 15 44 19

mvelkCisco:
I'm not sure if we can remove the `magnetic` prop or what's the purpose of keeping both atomic and magnetic version. @rwharrisjr?

rwharrisjr:
@mvelkCisco Originally the `magnetic` prop was to avoid breaking page layouts for projects doing a 1:1 swap from atomic-react to magna-react. For magnetic styles that affected sizes and potentially page layouts, we wanted to make it opt-in to avoid putting more work on developers to get a quick "magna-tization".

It's worth considering dropping that strategy now, as the initial migrations have mostly been done for the XDR projects. I still have some concern for other products that use atomic-react and will likely migrate to magna-react in the future. I don't know the timelines for that and what the situation will be with the magnetic team's full magnetic-react library.

In this specific case I think the size change is small enough and usage of the `ABreadcrumb` component is low enough that removing the `magnetic` prop is a low to no risk change.

rwharrisjr (approved):

PR #235, opened 2023-03-03
correction of displaced selected button in APagination component

rrubas:
Pull request solves the undesirable shift of the selected button by 2px to the left which occurred due to inheritance from the AButtonGroup component


rwharrisjr (approved):
nice fix!

PR #253, opened 2023-03-07
fix small AButton component

### LINKED ISSUE #250: Small button variant font-size is overwritten by `.a-app button`
**Describe the bug**

When using the `small` variant of `AButton`, the font-size shows 14px instead of 12px because of styles applied in `typography.scss`. It seems the style declaration needs to be more specific.

Screenshot 2023-03-06 at 10 28 37 PM

rrubas:
Closes #250


rwharrisjr (approved):
Pushed commit to do the same change to `--medium` for consistency.

PR #255, opened 2023-03-07
Add support for setting infoTooltip props

### LINKED ISSUE #254: Input infoTooltip not usable in drawers
Some drawers contain a lot of content and are set to scroll, and they also include some inputs. We have 2 options:

1. Use `AMount` so the tooltip mounts inside the scrollable content and it scrolls correctly. This makes the tooltip cut off:

image

2. Not use `AMount` so that the tooltip mounts outside the drawer to prevent cut off. This prevents the tooltip from scrolling:

https://user-images.githubusercontent.com/51229231/223455783-13220d6e-632d-4367-890b-ed46544ab31a.mov

I don't think we can have both at the same time:
- We either need to mount further up (braking scrolling): https://stackoverflow.com/a/40108007
- And setting overflow in one direction and scroll in other doesn't seem possible based on https://stackoverflow.com/a/38793442

mvelkCisco:
Fixes #254 by allowing me to set `placement` prop on the `infoTooltip`.

### DIFF
--- a/framework/components/ATextInput/ATextInput.js
+++ b/framework/components/ATextInput/ATextInput.js
@@ -7,6 +7,7 @@ import React, {
   useState
 } from "react";
 
+import ATooltip from "../ATooltip";
 import AInputBase from "../AInputBase";
 import {AFormContext} from "../AForm";
 import AIcon from "../AIcon";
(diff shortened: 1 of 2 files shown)


mvelkCisco:
To clarify, to work around the issue I would just align those couple of tooltips to the right.

rwharrisjr (approved):

framework/components/ATextInput/ATextInput.js
```diff
@@ -435,6 +436,14 @@ ATextInput.propTypes = {
...
    */
   label: PropTypes.node,
+  /**
+   * Adds an information tooltip next to the label
+   */
+  infoTooltip: PropTypes.string,
+  /**
+   * Overrides props of `ATooltip` used to display `infoTooltip`
+   */
+  infoTooltipProps: PropTypes.instanceOf(ATooltip.propTypes),
```
mvelkCisco, line 446: Also I don't think I can do this. It works fine for docs but not for actual prop validation.

mvelkCisco: I'll fix it in the next PR.

mvelkCisco: Should be fixed in #256.

PR #256, opened 2023-03-07
Tooltip delay proposal

mvelkCisco:
Delays showing/hiding the tooltip by default when toggling `open` prop. This can be breaking for apps that implement delays themselves (like I do in intelligence), but it would make everything more consistent. 

Also adds a tip by default.

Examples updated to show tooltip on hover which is a more common and recommended use case: https://magnetic.cisco.com/0a43ab5cd/p/727988-popover

Relates to #255.

### DIFF
--- a/framework/components/AFieldBase/AFieldBase.js
+++ b/framework/components/AFieldBase/AFieldBase.js
@@ -88,6 +88,8 @@ const AFieldBase = forwardRef(
   }
 );
 
+const {anchorRef, ...infoTooltipProps} = ATooltip.propTypes;
+
 AFieldBase.propTypes = {
   /**
    * Sets the hint content.
@@ -116,7 +118,7 @@ AFieldBase.propTypes = {
   /**
    * Overrides props of `ATooltip` used to display `infoTooltip`
    */
-  infoTooltipProps: PropTypes.instanceOf(ATooltip.propTypes),
+  infoTooltipProps: PropTypes.shape(infoTooltipProps),
   /**
    * Handles the label's `click` event.
    */
--- a/framework/components/AFieldBase/AFieldBase.js
+++ b/framework/components/AFieldBase/AFieldBase.js
@@ -116,7 +116,7 @@ AFieldBase.propTypes = {
    */
   infoTooltip: PropTypes.string,
   /**
-   * Overrides props of `ATooltip` used to display `infoTooltip`
+   * Overrides props of `ATooltip` used to display `infoTooltip`. See `ATooltip.propTypes`.
    */
   infoTooltipProps: PropTypes.shape(infoTooltipProps),
   /**
--- a/framework/components/AFieldBase/AFieldBase.js
+++ b/framework/components/AFieldBase/AFieldBase.js
@@ -3,7 +3,7 @@ import React, {forwardRef, useRef, useState} from "react";
 
 import AHint from "../AHint";
 import AIcon from "../AIcon";
-import ATooltip from "../ATooltip";
+import ATooltip, {ATooltipPropTypes} from "../ATooltip";
 import "./AFieldBase.scss";
 
 const AFieldBase = forwardRef(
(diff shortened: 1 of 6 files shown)


framework/components/ATooltip/ATooltip.js
```diff
@@ -14,15 +20,52 @@ const ATooltip = forwardRef(
...
+    }, [closeDelay, openDelay]);
+
+    // sync open prop change to internal state with delay
+    useEffect(() => {
+      if (open) {
+        openWithDelay();
+      } else {
+        closeWithDelay();
+      }
+    }, [open, openWithDelay, closeWithDelay]);
```
rwharrisjr, line 67: The `useEffect` needs a cleanup in case the tooltip unmounts before the timeout triggers

```
  return () => {
        if (timeout.current) {
        clearTimeout(timeout.current);
        }
      };
```

framework/components/ATooltip/ATooltip.mdx
```diff
@@ -143,7 +143,7 @@ The `ATooltip` component inherits passed props.
       </ARow>
       <ARow>
         <ACol className="text-right">
-          <AButton ref={buttonRef11} onClick={() => setOpen11(!open11)}>
+          <AButton ref={buttonRef11}  onMouseEnter={() => setOpen11(true)} onMouseLeave={() => setOpen11(false)}>
```
rwharrisjr, line 146: Nice, that has always bugged me that they weren't on hover.

rwharrisjr (changes requested):

framework/components/ATooltip/ATooltip.js
mvelkCisco: Right, good catch.

framework/components/AFieldBase/AFieldBase.js
```diff
@@ -114,9 +116,9 @@ AFieldBase.propTypes = {
    */
   infoTooltip: PropTypes.string,
   /**
-   * Overrides props of `ATooltip` used to display `infoTooltip`
+   * Overrides props of `ATooltip` used to display `infoTooltip`. See `ATooltip.propTypes`.
    */
-  infoTooltipProps: PropTypes.instanceOf(ATooltip.propTypes),
+  infoTooltipProps: PropTypes.shape(infoTooltipProps),
```
mvelkCisco, line 121: `instanceOf` renders nicely in docs but does not actually validate props correctly. This does not render nicely in docs but at least it works.

framework/components/ATooltip/ATooltip.js
```diff
@@ -51,7 +100,7 @@ const ATooltip = forwardRef(
   }
 );
 
-ATooltip.propTypes = {
+export const ATooltipPropTypes = {
```
mvelkCisco, line 103: Looks like docs build has issues accessing `ATooltip.propTypes`. Maybe it's stripped out due to production build. This is a workaround.

brennarvo:
This makes sense and is a nice enhancement. I'm just not sure how I feel about a `useEffect` that updates internal state based off props. React actually makes a note of this: https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes. We're essentially doing a double-render whenever `open` changes.

I think maybe we could extract this out into a hook. That way, it's also not a breaking change, and we can still have the `open` prop be the "source of truth" for when the tooltip is opened"

```js
const { isOpen, open, close } = useTooltipState({ delay: "400ms" });

return (
  <>
    <button onMouseEnter={open} onMouseLeave={close}>
      Open Tooltip
    </button>
    <ATooltip open={isOpen} />
  </>
)
```

rwharrisjr:
Using a pattern we have elsewhere, this solution removes the controlled+uncontrolled nature of the current proposal and moves logic to a hook. This also has the benefit of no direct state management for easier readability. https://github.com/cisco-sbg-ui/magna-react/pull/268.

Effectively the linked PR takes the function code from this PR, puts it in a hook, and leaves `ATooltip` as is.

mvelkCisco:
To be honest I wasn't super thrilled with this proposal either. But I didn't have a better idea at the time. Now I would rather continue with the direction and discussion in #268.

PR #268, opened 2023-03-08
Add useToggle hook for delay open of tooltip

rwharrisjr:
Adds `useToggle` as a hook for controlling components with an "open" prop. Provides

- isOpen state prop
- function to open by event
- function to close by event
- function to toggle by event

`const {isOpen, open, close} = useToggle(400, 0);`

### DIFF
--- a/framework/components/AFieldBase/AFieldBase.js
+++ b/framework/components/AFieldBase/AFieldBase.js
@@ -1,6 +1,7 @@
 import PropTypes from "prop-types";
 import React, {forwardRef, useRef, useState} from "react";
 
+import useToggle from "../../hooks/useToggle";
 import AHint from "../AHint";
 import AIcon from "../AIcon";
 import ATooltip from "../ATooltip";
@@ -25,7 +26,11 @@ const AFieldBase = forwardRef(
     ref
   ) => {
     const infoIconRef = useRef(),
-      [infoTooltipOpen, setInfoTooltipOpen] = useState(false);
+      {
+        isOpen: isInfoTooltipOpen,
+        open: openInfoTooltip,
+        close: closeInfoTooltip
+      } = useToggle();
 
     let className = "a-field-base";
 
@@ -55,15 +60,15 @@ const AFieldBase = forwardRef(
               <>
                 <AIcon
                   ref={infoIconRef}
-                  onMouseEnter={() => setInfoTooltipOpen(true)}
-                  onMouseLeave={() => setInfoTooltipOpen(false)}
+                  onMouseEnter={openInfoTooltip}
+                  onMouseLeave={closeInfoTooltip}
                 >
                   information
                 </AIcon>
                 <ATooltip
                   anchorRef={infoIconRef}
                   id={`${labelId}-tooltip`}
-                  open={infoTooltipOpen}
+                  open={isInfoTooltipOpen}
                   placement="top"
                   pointer
                   {...infoTooltipProps}
--- a/framework/components/AFieldBase/AFieldBase.js
+++ b/framework/components/AFieldBase/AFieldBase.js
@@ -1,10 +1,10 @@
 import PropTypes from "prop-types";
-import React, {forwardRef, useRef, useState} from "react";
+import React, {forwardRef} from "react";
 
-import useToggle from "../../hooks/useToggle";
 import AHint from "../AHint";
 import AIcon from "../AIcon";
 import ATooltip from "../ATooltip";
+import ATriggerTooltip from "../ATriggerTooltip";
 import "./AFieldBase.scss";
 
 const AFieldBase = forwardRef(
@@ -25,13 +25,6 @@ const AFieldBase = forwardRef(
     },
     ref
   ) => {
-    const infoIconRef = useRef(),
-      {
-        isOpen: isInfoTooltipOpen,
-        open: openInfoTooltip,
-        close: closeInfoTooltip
-      } = useToggle();
-
     let className = "a-field-base";
 
     if (validationState !== "default") {
@@ -57,25 +50,15 @@ const AFieldBase = forwardRef(
               <span className="a-field-base__label--required">*</span>
             )}
             {infoTooltip && (
-              <>
-                <AIcon
-                  ref={infoIconRef}
-                  onMouseEnter={openInfoTooltip}
-                  onMouseLeave={closeInfoTooltip}
-                >
-                  information
-                </AIcon>
-                <ATooltip
-                  anchorRef={infoIconRef}
-                  id={`${labelId}-tooltip`}
-                  open={isInfoTooltipOpen}
-                  placement="top"
-                  pointer
-                  {...infoTooltipProps}
-                >
-                  {infoTooltip}
-                </ATooltip>
-              </>
+              <ATriggerTooltip
+                id={`${labelId}-tooltip`}
+                placement="top"
+                content={infoTooltip}
+                pointer
+                {...infoTooltipProps}
+              >
+                <AIcon>information</AIcon>
+              </ATriggerTooltip>
             )}
           </label>
         )}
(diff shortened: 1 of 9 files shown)


mvelkCisco:
I agree `useEffect` delaying internal state change is nasty. This is more neat and basically what we do in https://github.com/advthreat/intelligence/blob/master/src/hooks/useTooltip.js. 

However it doesn't solve the original issue I had. My intention was to also delay this `infoTooltip` that's build in `AFieldBase` https://github.com/cisco-sbg-ui/magna-react/blob/canary/framework/components/AFieldBase/AFieldBase.js#L63.
Maybe we can use `useToggle` in `AFieldBase` and expose its args as a `AFieldBase` prop?

framework/components/ATooltip/ATooltip.mdx
```diff
@@ -16,7 +16,33 @@ import {ATooltip} from "@cisco-sbg-ui/magna-react";
 
 It is necessary for `ATooltip` to be a descendant of `AApp` for mounting.
 
-## Usage
+## Usage (hook with delay options)
+
+<Playground
+  code={`() => {
+  const iconRef = useRef(null);
+  const {isOpen, open, close} = useToggle(400, 0);
```
mvelkCisco, line 24: For the example I would loose the args of `useToggle` to encorage using defaults for consistency.

```suggestion
  const {isOpen, open, close} = useToggle();
```

framework/hooks/useToggle.js
```diff
@@ -0,0 +1,36 @@
+import {useCallback, useRef, useState} from "react";
+
+const useToggle = (openDelay = 400, closeDelay = 0) => {
```
mvelkCisco, line 3: `closeDelay` should default to `undefined` so that `openDelay / 2` fallback is used.

mvelkCisco:
Speaking of the tooltip, maybe including an example on how to write a wrapper for `ATooltip` (something like this https://github.com/advthreat/intelligence/blob/master/src/components/Tooltip.js#L41) would make the life of new comers and maybe also existing users easier. Or maybe just straight up export it as `AHoverTooltip` or something like that.

Don't get me wrong, I love the flexibility you are aiming for. But adding `useRef`, `useState` (now `useToggle`) and 2 event handlers for the tooltip target every single time is really not ergonomic (and eg in data table cell I would say not advised).

In 99% of cases I just want the tooltip to show and hide on hover, with a consistent delay. That can be as easy as `<AHoverTooltip content="foo">bar</AHoverTooltip>` with a wrapper component.

PS: My wrapper has some downsides, it wraps multiple children/number/string with a `<span>` to be able to attach `ref` and does not work with react function components that don't implement `forwardRef` (that can be fixed with an extra `<span>` too). In some case that `span` might cause css issues.

rwharrisjr:
I added `useTooltip` to `AFieldBase`. I think it's safe enough to make that standard behavior in this situation. 


I 100% agree with making `AHoverTooltip`.  The requirement for components using `forwardRef` in your implementation is pretty standard across tooltip libraries I've seen (@tippyjs/react). I will add a similar implementation in magna-react.

rwharrisjr:
@mvelkCisco Added `ATriggerTooltip` to this PR

Usage will be 

```
  <ATriggerTooltip
    id="tooltip_usage_1"
    placement="bottom-left"
    content="Tooltip text"
    pointer>
     <AIcon>
    information</AIcon>
  </ATooltip>
```

framework/hooks/useToggle.js
```diff
@@ -0,0 +1,40 @@
...
+      closeDelay ?? openDelay / 2
+    );
+  }, [closeDelay, openDelay, timeout]);
+
+  const toggle = useCallback(() => {
+    isOpen ? close() : open();
+  }, [isOpen, open, close]);
+
+  useEffect(() => {
+    timeout.current && clearTimeout(timeout.current);
```
brennarvo, line 34: Can you describe what this `useEffect` is for? It seems like whenever we change `timeout` in `useToggle`, we clear the timeout anyways, such as in the `open` and `close` functions.

If it's for a cleanup, it needs to be a returned function from `useEffect`

rwharrisjr: That is meant to be cleanup. My brain skipped the needing to return it when I added that.

framework/components/AFieldBase/AFieldBase.js
```diff
@@ -52,25 +50,15 @@ const AFieldBase = forwardRef(
...
-                  {...infoTooltipProps}
-                >
-                  {infoTooltip}
-                </ATooltip>
-              </>
+              <ATriggerTooltip
+                id={`${labelId}-tooltip`}
+                placement="top"
+                content={infoTooltip}
+                pointer
```
mvelkCisco, line 57: Can you please make `pointer` `true` a default in `ATooltip`? I don't think I've ever seen a tooltip without a pointer in magnetic. This is no-pointer version is coming from atomic.

rwharrisjr: Done

framework/components/ATriggerTooltip/ATriggerTooltip.js
```diff
@@ -0,0 +1,77 @@
...
+    );
+  });
+
+  return (
+    <>
+      {React.Children.map(children, (child, index) =>
+        React.cloneElement(child, {
+          ...child.props,
+          ref: (ref) => {
+            childrenRef.current[index] = ref;
```
brennarvo, line 55: We'll also want to pass the `ref` passed to this internal function to a possible `forwardRef` to the `ATriggerTooltip` children.

Like:

```jsx
function MyFeature() {
  const featureRef = useRef();

  return (
    <ATriggerTooltip>
      <p ref={featureRef}>I am the trigger for the tooltip component I am wrapped in.</p>
    </ATriggerTooltip>
  )
}
```
We'll want the `MyFeature` component to be able to access the `ref` on the `<p>` element.

```diff
ref: (node) {
  childrenRef.current[index] = ref;
+  if (typeof triggerTooltipForwardedRef === 'function') {
+  } else {
+   triggerTooltipForwardedRef.current = node;
+  }
}
```

framework/components/ATriggerTooltip/ATriggerTooltip.js
```diff
@@ -0,0 +1,77 @@
...
+  const childrenRef = useRef([]);
+  const {isOpen, open, close, toggle} = useToggle(openDelay, closeDelay);
+
+  useEffect(() => {
+    const childRefs = childrenRef.current;
+    switch (trigger) {
+      case "mouseenter":
+        childRefs.forEach((childRef) => {
+          childRef.addEventListener("mouseenter", open);
+          childRef.addEventListener("mouseleave", close);
```
brennarvo, line 24: Is there a reason for not adding these event listeners where we clone the child, i.e.,:

```js
return (
  <>
    {React.Children.map(children, (child, index) =>
      React.cloneElement(child, {
        ...child.props,
        onMouseEnter: (e) => {
          open();
          const childOnMouseEnter = child.props.onMouseEnter;
          childOnMouseEnter && childOnMouseEnter(e);
        },
        onMouseLeave: (e) => {
          close();
          const childOnMouseLeave = child.props.onMouseLeave;
          childOnMouseLeave && childOnMouseLeave(e);
        },
        onClick: (e) => {
          open();
          const childOnClick = child.props.onClick;
          childOnClick && childOnClick(e);
        },
        ref: (ref) => {
          childrenRef.current[index] = ref;
        },
      })
    )}
    {childElements}
  </>
)
```

framework/components/ATriggerTooltip/ATriggerTooltip.js
brennarvo: Or, if you want to use the utility in Magna React:

```diff
+ ref: handleMultipleRefs(props.ref, (node) => {
+   childrenRef.current[index] = node;
+ })
```

brennarvo (approved):
👍

PR #275, opened 2023-03-09
Abstract menu placement into a hook; use hook in ASelect and ACombobox

rwharrisjr:
Adds bottom edge detection to ACombobox. Logic has been abstracted into the `useMenuSpacing` hook. 

This should potentially be enhanced to add scroll detection to recalculate the menu placement.

### DIFF
--- a/framework/components/AMenuBase/hooks.js
+++ b/framework/components/AMenuBase/hooks.js
@@ -1,0 +1,34 @@
+import {useCallback, useState} from "react";
+
+/**
+ * Determines if a menu should flip its placement to above an
+ * element if there is not enough vertical space below to
+ * render the entire menu.
+ *
+ * To be used with ASelect and other components that use AMenu
+ * in a dropdown sort of way.
+ */
+
+const useMenuSpacing = (surfaceRef, menuRef, maxHeight) => {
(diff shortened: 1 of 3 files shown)


ericamann (approved):
🚀

framework/components/AMenuBase/hooks.js
```diff
@@ -0,0 +1,34 @@
...
+ * in a dropdown sort of way.
+ */
+
+const useMenuSpacing = (surfaceRef, menuRef, maxHeight) => {
+  const [menuSpace, setMenuSpace] = useState(0);
+
+  const checkMenuSpacing = useCallback(() => {
+    const inputBottom = surfaceRef?.current?.getBoundingClientRect().bottom,
+      verticalSpace = inputBottom
+        ? window.innerHeight - inputBottom - 10
```
brennarvo, line 18: I know `10` was already being used in `ASelect`, but any idea what that represents? Is it to account for the actual `<select>` element the dropdowns render underneath?

rwharrisjr: I think it's just supposed to be a buffer for good measure.

PR #280, opened 2023-03-10
AApp/ATheme listen for theme prop changes to switch theme

### LINKED ISSUE #279: Changing a theme on AApp does not get reflected in themeContext
All of the MFE apps are getting a `theme` prop from the shell app, which is then passed to the `AApp` as `theme` prop to sync the intelligence app theme when theme switch is flipped in the shell app.

However `AApp` does not have a `theme` prop. It does have `defaultTheme` prop but changing that is probably not reflected in the context: https://github.com/cisco-sbg-ui/magna-react/blob/2816356ae16660d31c4aa7ded60958f22e9465bf/framework/components/ATheme/ATheme.js#L18

Maybe we should add a `theme` prop to `AApp` -> `ATheme`?

https://user-images.githubusercontent.com/12389266/224390230-0db636fc-dcaf-4768-9d0b-1edd6885fc45.mov

mvelkCisco:
Adds support for `theme` prop to `ATheme` and `AApp` that can be used to sync nested app theme with its parent's theme. 

Some apps are using `theme`, assuming the theme will get propagated from the parent app to their nested app. Eg:
- https://github.com/advthreat/incident-manager/blob/master/src/AppWrapper.js#L9
- https://github.com/advthreat/intelligence/blob/master/src/module-federation/RemoteApp.js#L63

This has no effect as the `theme` prop does not exist.

These apps could use `defaultTheme` instead, but changes to `defaultTheme` prop are not reflected back to the current theme (which given its name makes sense).

At least some apps have figured this out and use `useEffect` to call `setCurrentTheme` to sync its theme: https://github.com/advthreat/securex-ui-monorepo/pull/304/files#diff-96ad99e94990287ef61dc6a42d6124b9cf297d63c9e2f8a4f245d81a727dc149R117
However that seems like a very convoluted way of doing things. This `useEffect` also needs to be nested 1 layer deeper than the app's `AApp`.

`ATheme` has been cleaned up a bit in this PR. I think there were some logic mistakes too.

fixes #279

### DIFF
--- a/framework/components/AApp/AApp.js
+++ b/framework/components/AApp/AApp.js
@@ -68,7 +68,7 @@ AApp.propTypes = {
    */
   defaultTheme: PropTypes.oneOf(["default", "dusk"]),
   /**
-   * Sets the current theme. Changes to this prop are reflected as a current theme. Takes precedence over defaultTheme. Do not use "theme" and "persist" props at the same time. Providing a "theme" prop indicates that the theme is managed from outside.
+   * Sets the current theme. Changes to this prop are reflected as a current theme. Takes precedence over defaultTheme. Do not use "theme" and "persist" props at the same time. Providing a "theme" prop indicates that the theme is managed from outside and it will not be persisted.
    */
   theme: PropTypes.oneOf(["default", "dusk"])
 };
--- a/framework/components/AApp/AApp.js
+++ b/framework/components/AApp/AApp.js
@@ -11,6 +11,7 @@ const AApp = forwardRef(
       animations = true,
       children,
       className: propsClassName,
+      theme,
       defaultTheme,
       persistTheme = false,
       scrollbars = true,
@@ -40,7 +41,9 @@ const AApp = forwardRef(
         component={ATheme}
         wrapClassName="a-app--wrap"
         persist={persistTheme}
-        defaultTheme={defaultTheme}>
+        theme={theme}
+        defaultTheme={defaultTheme}
+      >
         {children}
       </AMount>
     );
@@ -53,17 +56,21 @@ AApp.propTypes = {
    */
   animations: PropTypes.bool,
   /**
-   * Sets the default theme.
+   * Toggles styled scrollbars.
    */
-  defaultTheme: PropTypes.oneOf(["default", "dusk"]),
+  scrollbars: PropTypes.bool,
   /**
-   * Toggles whether the theme is persisted in local storage.
+   * Toggles whether the theme is loaded from local storage on mount, and persisted to local storage on theme change.
    */
   persistTheme: PropTypes.bool,
   /**
-   * Toggles styled scrollbars.
+   * Sets the default theme on mount. Changes to this prop are not reflected as a current theme.
    */
-  scrollbars: PropTypes.bool
+  defaultTheme: PropTypes.oneOf(["default", "dusk"]),
+  /**
+   * Sets the current theme. Changes to this prop are reflected as a current theme. Takes precedence over defaultTheme.
+   */
+  theme: PropTypes.oneOf(["default", "dusk"])
 };
 
 AApp.displayName = "AApp";
--- a/framework/components/AApp/AApp.js
+++ b/framework/components/AApp/AApp.js
@@ -68,7 +68,7 @@ AApp.propTypes = {
    */
   defaultTheme: PropTypes.oneOf(["default", "dusk"]),
   /**
-   * Sets the current theme. Changes to this prop are reflected as a current theme. Takes precedence over defaultTheme.
+   * Sets the current theme. Changes to this prop are reflected as a current theme. Takes precedence over defaultTheme. Do not use "theme" and "persist" props at the same time. Providing a "theme" prop indicates that the theme is managed from outside.
    */
   theme: PropTypes.oneOf(["default", "dusk"])
 };
--- a/framework/components/ATheme/ATheme.js
+++ b/framework/components/ATheme/ATheme.js
@@ -12,14 +12,15 @@ const ATheme = forwardRef(
     ref
   ) => {
     const [currentTheme, setCurrentTheme] = useState("dusk");
-    const isDark = currentTheme === "dusk"
-    const isLight = currentTheme !== "dusk"
+    const isDark = currentTheme === "dusk";
+    const isLight = currentTheme !== "dusk";
 
     useIsomorphicLayoutEffect(() => {
       let initialTheme = "default";
       if (persist) {
         if (Object.prototype.hasOwnProperty.call(localStorage, LS_KEY)) {
-          initialTheme = localStorage.getItem(LS_KEY) === "dusk" ? "dusk" : "default";
+          initialTheme =
+            localStorage.getItem(LS_KEY) === "dusk" ? "dusk" : "default";
         } else if (["default", "dusk"].includes(defaultTheme)) {
           initialTheme = defaultTheme;
         } else if (
@@ -33,7 +34,8 @@ const ATheme = forwardRef(
       }
 
       setCurrentTheme(initialTheme);
-    }, []); // eslint-disable-line react-hooks/exhaustive-deps
+      // listen for defaultTheme prop change
+    }, [defaultTheme]);
 
     const themeContext = {
       persist,
--- a/framework/components/ATheme/ATheme.js
+++ b/framework/components/ATheme/ATheme.js
@@ -39,6 +39,8 @@ const ATheme = forwardRef(
     // eagerly returns the highest priority setting
     const getInitialClientTheme = useCallback(() => {
       // supported theme in 'theme' prop overrides 'defaultTheme' and ignores 'persist' flag
+      // (this being the first case also works around a race condition between the useEffect initializing the currentTheme
+      // and the useEffect syncing the currentTheme with theme prop by both having the same outcome)
       if (isSupportedTheme(theme)) {
         return theme;
       }
--- a/framework/components/ATheme/ATheme.js
+++ b/framework/components/ATheme/ATheme.js
@@ -30,24 +30,26 @@ const ATheme = forwardRef(
   ) => {
     const [currentTheme, setCurrentTheme] = useState(DEFAULT_INITIAL_THEME);
 
+    // eagerly returns the highest priority setting
     const getInitialClientTheme = useCallback(() => {
-      // take initial theme from 'defaultTheme' prop or fallback to DEFAULT_THEME
-      let initialTheme = isSupportedTheme(defaultTheme)
-        ? defaultTheme
-        : DEFAULT_INITIAL_THEME;
-      // supported theme in 'theme' prop overrides 'defaultTheme'
-      if (isSupportedTheme(theme)) {
-        initialTheme = theme;
-      }
-      // when persist is enabled and available
+      // persist is enabled and localStorage available
       if (persist && typeof localStorage !== "undefined") {
-        const persistedTheme = localStorage.getItem(LS_KEY);
         // supported persisted theme overrides both 'defaultTheme' and 'theme' props
+        const persistedTheme = localStorage.getItem(LS_KEY);
         if (isSupportedTheme(persistedTheme)) {
-          initialTheme = persistedTheme;
+          return persistedTheme;
         }
       }
-      return initialTheme;
+      // supported theme in 'theme' prop overrides 'defaultTheme'
+      if (isSupportedTheme(theme)) {
+        return theme;
+      }
+      // supported theme in 'defaultTheme'
+      if (isSupportedTheme(defaultTheme)) {
+        return defaultTheme;
+      }
+      // fallback
+      return DEFAULT_INITIAL_THEME;
     }, [persist, theme, defaultTheme]);
 
     // set initial theme based on client settings (AAutoTheme can override this further)
--- a/framework/components/ATheme/ATheme.js
+++ b/framework/components/ATheme/ATheme.js
@@ -15,18 +15,18 @@ function isSupportedTheme(theme) {
 }
 
 class ThemeStorage {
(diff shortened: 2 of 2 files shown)


framework/components/ATheme/ATheme.js
```diff
@@ -1,55 +1,90 @@
...
-    useIsomorphicLayoutEffect(() => {
-      let initialTheme = "default";
-      if (persist) {
-        if (Object.prototype.hasOwnProperty.call(localStorage, LS_KEY)) {
-          initialTheme = localStorage.getItem(LS_KEY) === "dusk" ? "dusk" : "default";
-        } else if (["default", "dusk"].includes(defaultTheme)) {
-          initialTheme = defaultTheme;
-        } else if (
-          window.matchMedia &&
-          window.matchMedia("(prefers-color-scheme: dark)").matches
```
mvelkCisco, line 27: I think this is supposed to be handled by `AAutoTheme` and I'm not even sure how this behaved.

framework/components/ATheme/ATheme.js
```diff
@@ -1,55 +1,90 @@
...
+    },
     ref
   ) => {
-    const [currentTheme, setCurrentTheme] = useState("dusk");
-    const isDark = currentTheme === "dusk"
-    const isLight = currentTheme !== "dusk"
+    const [currentTheme, setCurrentTheme] = useState(DEFAULT_THEME);
 
-    useIsomorphicLayoutEffect(() => {
-      let initialTheme = "default";
```
mvelkCisco, line 19: This is super confusing because there's `useState("dusk")` but `let initialTheme = "default"`.
This flips the current theme to `default` in this effect, but only if `persist` is true. So with `persist` the default was `dusk`, without `persist` it was `default`. 

In the new implementation, I assume `default` should be the default regardless the `persist` flag.

framework/components/AApp/AApp.js
```diff
@@ -53,17 +56,21 @@ AApp.propTypes = {
...
   /**
-   * Toggles styled scrollbars.
+   * Sets the default theme on mount. Changes to this prop are not reflected as a current theme.
+   */
+  defaultTheme: PropTypes.oneOf(["default", "dusk"]),
+  /**
+   * Sets the current theme. Changes to this prop are reflected as a current theme. Takes precedence over defaultTheme.
    */
-  scrollbars: PropTypes.bool
+  theme: PropTypes.oneOf(["default", "dusk"])
```
mvelkCisco, line 73: I was hoping I could import `SUPPORTED_THEMES` from `ATheme` here, but the SSR breaks down when I do. It just sees a sting `SUPPORTED_THEMES` in `oneOf`.

framework/components/ATheme/ATheme.js
```diff
@@ -1,55 +1,90 @@
...
-        if (Object.prototype.hasOwnProperty.call(localStorage, LS_KEY)) {
-          initialTheme = localStorage.getItem(LS_KEY) === "dusk" ? "dusk" : "default";
-        } else if (["default", "dusk"].includes(defaultTheme)) {
-          initialTheme = defaultTheme;
-        } else if (
-          window.matchMedia &&
-          window.matchMedia("(prefers-color-scheme: dark)").matches
-        ) {
-          initialTheme = "dusk";
+    const getInitialClientTheme = useCallback(() => {
```
MikkCZ, line 31: This is only called from `useEffect`, which is only triggered once. Could you inline it there?

framework/components/ATheme/ATheme.js
```diff
@@ -1,55 +1,90 @@
...
+import React, {forwardRef, useState, useCallback} from "react";
 
 import {useIsomorphicLayoutEffect} from "../../utils/hooks";
 import AThemeContext from "./AThemeContext";
 
 const LS_KEY = "persist-magna-react-theme";
 
+const DEFAULT_THEME = "default";
+const DUSK_THEME = "dusk";
+const SUPPORTED_THEMES = [DEFAULT_THEME, DUSK_THEME];
```
MikkCZ, line 11: I would just do
```js
const SUPPORTED_THEMES = {
  DEFAULT: "default",
  DUSK: "dusk"
};
```

If you need to iterate over them somewhere, use `Object.values(SUPPORTED_THEMES)`.

framework/components/ATheme/ATheme.js
MikkCZ: I guess this is supposed to automatically follow the system/browser settings.

MikkCZ (changes requested):

framework/components/ATheme/ATheme.js
```diff
@@ -1,55 +1,90 @@
...
-          initialTheme = "dusk";
+    const getInitialClientTheme = useCallback(() => {
+      // take initial theme from 'defaultTheme' prop or fallback to DEFAULT_THEME
+      let initialTheme = isSupportedTheme(defaultTheme)
+        ? defaultTheme
+        : DEFAULT_THEME;
+      // supported theme in 'theme' prop overrides 'defaultTheme'
+      if (isSupportedTheme(theme)) {
+        initialTheme = theme;
+      }
```
MikkCZ, line 39: Also I would probably structure this code the other way around to eagerly return the first value available, or proceed to other fallbacks in order
```js
if (persist && typeof localStorage !== "undefined" && isSupportedTheme(localStorage.getItem(LS_KEY))) {
   return localStorage.getItem(LS_KEY);
} else if (isSupportedTheme(theme)) {
   return theme;
} else if (...) {
   ...
} else {
   return DEFAULT_THEME;
}
```

framework/components/ATheme/ATheme.js
```diff
@@ -1,55 +1,90 @@
...
+      setCurrentTheme(getInitialClientTheme());
+      // run just once
     }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
+    // reflect theme prop changes to currentTheme
+    useIsomorphicLayoutEffect(() => {
+      if (isSupportedTheme(theme)) {
+        setCurrentTheme(theme);
+      }
+    }, [theme]);
```
MikkCZ, line 62: Won't there be a race-condition for these two?

MikkCZ (changes requested):

framework/components/ATheme/ATheme.js
mvelkCisco: There a different component called `AAutoTheme` which sets the theme based on the system settings.

framework/components/ATheme/ATheme.js
mvelkCisco: Technically I could, but I separated it for better semantics.

MikkCZ: Not sure if there is any semantic difference, if you actually never care about the reference of `getInitialClientTheme`.

framework/components/ATheme/ATheme.js
mvelkCisco: I mostly wanted to have a `DEFAULT_THEME` const, so I can use it in both `useState` and as a fallback determining the initial theme, so they are kept in sync. But now I see the meaning is a bit confusing when one of the themes is also called `default`. I'll introduce a separate const for the syncing purpose.

Converting this to object just makes these harder to use in like 5 places. I never need the object. It's just 2 consts, I would keep it as it is for convenient access.

MikkCZ: Well, the idea would be to actually have an enum, and an object is probably as close as we could get to it here.

framework/components/ATheme/ATheme.js
mvelkCisco: Right, my bad. I'll reverse this.

framework/components/ATheme/ATheme.js
MikkCZ: Please add a comment to the body of `getInitialClientTheme`, that the preference from `theme` really matters also to prevent this race condition.

framework/components/ATheme/ATheme.js
mvelkCisco: To me this is just much easier to read.

I see a 3 line `useEffect` without dependencies that calls one function => in my brain this function runs on mount. Very explicit. I also like to double confirm with a comment.

I see a large `useEffect` without dependencies that references 3 different things => in my brain this is potentially a mistake that needs analysing. And I hate doing that :)

And yes, I don't care about the the reference of `getInitialClientTheme`. I could just remove `useCallback`, but I kind of add it automatically to avoid instantiating a function over and over again. It's possible it's a bigger performance overhead this way, not sure. But again, it stops my brain from immediately screaming "Missing useCallback!".

framework/components/ATheme/ATheme.js
```diff
@@ -1,55 +1,101 @@
...
+
+    // persist currentTheme on change
+    useIsomorphicLayoutEffect(() => {
+      // when "persist" is enabled, localStorage is supported, and "theme" is not managed from outside
+      if (persist && typeof localStorage !== "undefined" && !theme) {
+        localStorage?.setItem(LS_KEY, currentTheme);
+      }
+    }, [currentTheme, persist, theme]);
```
MikkCZ, line 80: This is the second place where `persist` and `theme` insteract or have some priority over each other. What about processing them once as

```js
const themeLocalStorage  = useMemo(() => {
   if (!theme && persist) {
     return localStorage;
   } else {
     return undefined;
   }
}, [theme, persist]);
```

And then call `themeStorage?.getItem`/`themeStorage?.setItem` without the need to repeatadly check whether to do it, or not.

Or even

```js
const themeLocalStorage  = useMemo(() => {
   if (!theme && persist) {
     return {
       get: () => localStorage?.getItem(LS_KEY) ?? undefined; // or coalesce to null, to mimic getItem call when the value is not there
       set: (t) => localStorage?.setItem(LS_KEY, t);
     };
   } else {
     return {
       get: () => undefined; // or null, to mimic getItem call when the value is not there
       set: (t) => {};
     };
   }
}, [theme, persist]);
```

mvelkCisco: Hm, good idea... I'll try to do something like this. I might need to sprinkle in `typeof localStorage !== "undefined"` here and there too. Docs were crashing with just optional chaining, not sure why. And maybe a more descriptive name would be nice.

framework/components/ATheme/ATheme.js
mvelkCisco: This morning I've realised that this was probably meant to provide the system theme detection by default and `AAutoTheme` then overrides this and adds controls to the detection. So I'll add it back here.

framework/components/ATheme/ATheme.js
mvelkCisco: Please see the latest version. I've abstracted the storage, but I've kept the decision weather to actually use it separate. Making the storage return `undefined` when `theme` is set seem too non-transparent to me. I do make the decision about using the storage in one place now though.

framework/components/ATheme/ATheme.js
mvelkCisco: I've ended up making the state `undefined` and I don't change it in the `theme` syncing hook util it has been initialized.

framework/components/ATheme/ATheme.js
```diff
@@ -16,6 +14,22 @@ function isSupportedTheme(theme) {
   return SUPPORTED_THEMES.includes(theme);
 }
 
+class ThemeStorage {
+  LS_KEY = "persist-magna-react-theme";
```
MikkCZ, line 18: `LS_KEY` can be `private readonly` inside the class.

MikkCZ (changes requested):

framework/components/ATheme/ATheme.js
mvelkCisco: That's only supported by typescript.

MikkCZ (commented):
Consider this approved from my side. But I will leave the actual approval for @rwharrisjr  or @brennarvo .

framework/components/ATheme/ATheme.js
```diff
@@ -1,55 +1,123 @@
...
+const SUPPORTED_THEMES = [DEFAULT_THEME, DUSK_THEME];
+
+const DEFAULT_INITIAL_THEME = DEFAULT_THEME;
+
+function isSupportedTheme(theme) {
+  return SUPPORTED_THEMES.includes(theme);
+}
+
+class ThemeStorage {
+  LS_KEY = "persist-magna-react-theme";
```
rwharrisjr, line 18: This seems to set the key in local storage to `undefined`. Removing the that key/value from local storage then seems to cause issues.

The following seems to resolve that on my machine:

```
const LS_KEY = "persist-magna-react-theme";

class ThemeStorage {
  static isSupported() {
    return typeof localStorage !== "undefined";
  }

  static loadTheme() {
    return localStorage.getItem(LS_KEY);
  }

  static saveTheme(theme) {
    localStorage.setItem(LS_KEY, theme);
  }
}
```

rwharrisjr (changes requested):

framework/components/ATheme/ATheme.js
MikkCZ: Good catch. The problem is the functions are static, and also what gets to `this` inside of them. @mvelkCisco there should be no need to actually move `LS_KEY` outside of the class, but you should be able to use it as `ThemeStorage.LS_KEY` (instead of `this.LS_KEY`).

mvelkCisco: Ups, I've missed that. 

In any case, `this.LS_KEY` works fine as long as `LS_KEY` is `static` too. And `LS_KEY` needs to be `static` anyway to be used as `ThemeStorage.LS_KEY`. Thinking about `this` in `static` class hurts my brain. `ThemeStorage.LS_KEY` is probably better anyway, so I'm going with that for now.

rwharrisjr: @MikkCZ yep that fixed it

rwharrisjr (approved):

PR #284, opened 2023-03-10
feat: implement roving focus in useFocusTrap

### LINKED ISSUE #283: Implementing roving tab focus in `useFocusTrap`
The [ARIA Authoring Practices Guide (APG)](https://www.w3.org/WAI/ARIA/apg/) mentions roving tab focus behavior, which `useFocusTrap` does not currently support.

From APG:

> In the following description, the term tabbable element refers to any element with a tabindex value of zero or greater. Note that values greater than 0 are strongly discouraged.
>
> When a dialog opens, focus moves to an element inside the dialog. See notes below regarding initial focus placement.
> * Tab:
>    * Moves focus to the next tabbable element inside the dialog.
>    * If focus is on the last tabbable element inside the dialog, moves focus to the first tabbable element inside the dialog.
> * Shift + Tab:
>      * Moves focus to the previous tabbable element inside the dialog.
>      * If focus is on the first tabbable element inside the dialog, moves focus to the last tabbable element inside the dialog.
> * Escape: Closes the dialog.

brennarvo:
Closes #283 by updating the `useFocusTrap` [`TreeWalker` instance](https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker) to accept the root (parent) node of the trap. This allows the focus trap to determine its first and last child, which allows for "roving" behavior.

This PR also fixes an issue related to the focus trap not working in a Modal due to the strict `e.stopPropagation()` behavior on `keydown` events inside that component which was merged in #244.

For a demonstration, navigate to [this preview branch's `useFocusTrap` docs](https://magna-react-git-feat-roving-focus-trap.securex-preview.app/hooks/use-focus-trap#usage) and constantly press either <kbd>Tab</kbd> or <kbd>Shift + Tab</kbd>. Focus will circulate through the trap.

### DIFF
--- a/framework/components/AModal/AModal.js
+++ b/framework/components/AModal/AModal.js
@@ -8,10 +8,26 @@ import useFocusTrap from "../../hooks/useFocusTrap/useFocusTrap";
 import {
   allowBodyScroll,
   handleMultipleRefs,
+  isBackwardTab,
+  isForwardTab,
   preventBodyScroll
 } from "../../utils/helpers";
 
 import "./AModal.scss";
+
+/**
+ * Reasons for not stopping propagation for the following keys:
+ *  - Escape Key: allows user to close the modal
+ *  - Forward and Backward Tab: allows the event
+ *    to bubble to the document, which is needed
+ *    since the modal's focus trap is registered
+ *    on the window document.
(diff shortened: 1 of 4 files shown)


rwharrisjr (approved):
very cool

PR #298, opened 2023-03-15
Close drawer/#292

### LINKED ISSUE #292: Add close icon button to ADrawer
Header title overflows Icon.  Drawer should provide a close icon option.
https://github.com/advthreat/incident-manager/issues/334

androche:
Closes #292 - Add close button option to drawer.  

- Adds options for close icon or a title
- Skipped className as usage of this is unlikely at this time.
- Adjustment on `ADrawerContent` padding.  Most repos are overwriting the `24px` with `padding: 16px 24px` so if you can't beat em'--

**Button with icon**
Screenshot 2023-03-15 at 11 31 39 AM

**Button with title**
Screenshot 2023-03-15 at 10 35 16 AM

**Functionality**

drawerClose

### DIFF
--- a/framework/components/ADrawer/ADrawer.js
+++ b/framework/components/ADrawer/ADrawer.js
@@ -2,6 +2,8 @@ import React, {forwardRef} from "react";
 import PropTypes from "prop-types";
 
 import AModal from "../AModal/AModal";
+import AButton from "../AButton/AButton";
+import AIcon from "../AIcon/AIcon";
 
 import "./ADrawer.scss";
 
@@ -20,6 +22,8 @@ const ADrawer = forwardRef(
       slim = false,
       slimHeight,
       slimWidth,
+      onClose,
+      closeTitle,
       style: propsStyle,
       ...rest
     },
@@ -32,6 +36,8 @@ const ADrawer = forwardRef(
     const orientation =
       slideIn === "bottom" || slideIn === "top" ? "horizontal" : "vertical";
     let className = `a-drawer a-drawer--${orientation} a-drawer--${position} a-drawer--${slideIn}`;
+    const closeButtonClassName = "a-drawer--close-button";
+
     const style = {...propsStyle};
 
     if (slim) {
@@ -82,6 +88,16 @@ const ADrawer = forwardRef(
           className={className}
           style={style}
         >
+          {onClose && (
+            <AButton
+              className={closeButtonClassName}
+              onClick={onClose}
+              icon
+              tertiaryAlt
+            >
+              {closeTitle ? closeTitle : <AIcon>x</AIcon>}
+            </AButton>
+          )}
           {children}
         </DrawerPanelComponent>
       );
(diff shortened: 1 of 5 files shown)


rwharrisjr (approved):

framework/components/ADrawer/ADrawer.js
```diff
@@ -20,6 +22,9 @@ const ADrawer = forwardRef(
       slim = false,
       slimHeight,
       slimWidth,
+      onClose,
```
mvelkCisco, line 25: I may be a bit late to the party but maybe something like `closeBtnOnClick` might have been a more accurate name for this.

mvelkCisco: Just to elaborate `onClose` indicates to me that it's a callback called when drawer is closing. But the drawer doesn't manage it's own state so `onClose` seems a bit confusing. At the same time `closeBtnProps` was added so the API is a bit inconstent.

framework/components/ADrawer/ADrawer.js
```diff
@@ -20,6 +22,9 @@ const ADrawer = forwardRef(
...
+      closeTitle,
```
mvelkCisco, line 26: Maybe `closeBtnChildren` would have been more descriptive. Or `closeBtnIcon`. At the same time I don't see it used in https://github.com/advthreat/incident-manager/pull/542/files so was it really needed?

framework/components/ADrawer/ADrawer.js
androche: Not too late - I'll add it to another PR!

framework/components/ADrawer/ADrawer.js
androche: For closeBtnProps or title? `CloseBtnProps` is used and is passing a closeTestId. Title is just for the option of a string. It wasn't required, but figured it wouldn't hurt to have the option of whatever ppl might use e.g., Close, Exit.

mvelkCisco: I like the `closeBtnProps`. I've meant `closeTitle`. If you don't plan to extensively use it, I would not add it. It increases the complexity, and for special cases it is still possible just not to pass that `closeBtnOnClick` callback and render whatever elements/buttons you want from outside as it was done before.

Also, at some point somebody will want to add an icon to that button next to the `closeTitle` string, or a tooltip for it... and the string prop just won't cut it. If you really want to provide the highest flexibility, I would make `closeTitle` into a slot (eg `closeBtnContent`) and allow passing any react component (`PropTypes.node` maybe?). (I would maybe get rid of `<div className="pa-2">` in that case as it can interact with the passed content).

In any case, this is not a big deal and feel free to keep it as it is. Just some observations/suggestions :)

androche: @mvelkCisco  Its much appreciate and all really good points! Now you that you mention it, I think anything in past was in fact a button with an icon and text next to it which IIRC is an option already handled in AIcon.   Sounds like its best to pull it and let them pass anything.  We had to get a build out for this one, but there's more behind the scenes work planned for Drawer 👍

PR #303, opened 2023-03-17
refactor: btn to absolute

androche:
Had to move to absolute due to need for flex column display coming from Apps.  Unfortunately, this means content on left will overflow again (which is what we were trying to solve for) so this means overflow will need to be managed in the element next to it here and there. Ideally it would be nice to have option to control header content and its children with ADrawerHeader, but the refactor necessary for that to happen in the apps would be quite a rewrite on their part.


androche:
(Needing build for https://github.com/advthreat/incident-manager/pull/542)

PR #308, opened 2023-03-20
#307 - changed disabled prop to ariaDisabled

### LINKED ISSUE #307: Allow tooltips on disabled buttons

We've been asked to add tooltips to many of our disabled buttons to tell the user why it is disabled. However, `AButton` disables mouse events on disabled buttons, so the tooltip does not work.

`AButton` should allow tooltips to be shown on hover when disabled.

I've tried attaching tooltips to a span that wraps the disabled button. However, this seems to not work always work in Chrome. The tooltip will show, but then it will not remove the tooltip `onMouseLeave`.

Example of implementation: https://github.com/advthreat/incident-manager/blob/master/src/features/incidentDetail/response/SelectEntityButton.js#L56

Since this is functionality that we are constantly being asked to implement, it would be best to just allow tooltips on disabled buttons.

rwharrisjr:
This changes the `disabled` prop on `AButton` to `ariaDisable`.

Behavior that relies on the `disabled` prop will be broken, notably form submission. Given buttons don't submit values to forms unless clicked, this is likely not an issue for this library. More information https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled

This will allow for tooltips on disabled `AButton` components.

Resolves #307

### DIFF
--- a/framework/components/AButton/AButton.js
+++ b/framework/components/AButton/AButton.js
@@ -96,13 +96,13 @@ const AButton = forwardRef(
       }
     } else if (component) {
       TagName = component;
-      props.disabled = disabled;
       props.value = value;
       props.target = target;
+      props["aria-disable"] = disabled;
     } else {
-      props.disabled = disabled;
       props.type = type;
       props.value = value;
+      props["aria-disabled"] = disabled;
     }
 
     return <TagName {...props}>{children}</TagName>;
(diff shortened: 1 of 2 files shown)


rwharrisjr:
Decided against this.

PR #314, opened 2023-03-21
Add ALoader component

### LINKED ISSUE #305: Add Magnetic tiny loader variants

Magna-react only provides `ADotLoader` to be used as a tiny loader for situations where a small loader needs to show (e.g. a table cell is updating, file is uploading, etc.) However, `ADotLoader` is not a part of Magnetic design. 

Magnetic shows tiny spinner icons and tiny progress loaders for these situations. Both `APageLoader` and `AProgressBar` don't provide these size variants. 

Additionally, `AProgressBar` doesn't have the ability to be used as an `indeterminate` progress bar.

**Example of indeterminate functionality:** https://mui.com/material-ui/react-progress/#linear-indeterminate

Components to be created for Magnetic tiny loaders. We also need an indeterminate progress bar. It looks like Magnetic has 3 size variants for tiny loaders:

Screenshot 2023-03-17 at 12 21 46 PM

https://www.figma.com/file/oVZWatImEIbl1c8sjdGxi0/branch/yU3qYkj6T2kfN4XYtUkEeD/%F0%9F%A7%B2--Magnetic-Design-Library?node-id=33305-356318&t=HcH2DgmjJ1JKgLep-0

rwharrisjr:
- Adds size variants to `ADotLoader`
- Adds new component `ALoader`

`ALoader` takes an indicator type, and allows for children. The children can be placed around the indicator using `childPlacement`.

Resolves #305 

----
##### Notes

This removes the `childPlacement` prop from `ASpinLoader` and moves it to `ALoader` to add the convenience component to swap out indicator types as needed. `ASpinLoader` has not been in a build yet, so this is a non-breaking change.

### DIFF
--- a/framework/components/ALoader/ALoader.js
+++ b/framework/components/ALoader/ALoader.js
@@ -4,7 +4,7 @@ import ADotLoader from "./ADotLoader";
 import ASpinLoader from "./ASpinLoader";
 import AProgressbar from "../AProgressbar";
 
-import "./AGenericLoader.scss";
+import "./ALoader.scss";
 
 const baseClass = "a-generic-loader";
 
@@ -14,14 +14,14 @@ export const INDICATORS = {
   dot: ADotLoader
 };
 
-const AGenericLoader = forwardRef(
+const ALoader = forwardRef(
   (
     {
       className: propsClassName,
       component = INDICATORS.spinner,
       indicatorProps = {},
       size,
-      childPlacement = "bottom",
+      placement = "top",
       children,
       ...rest
     },
@@ -51,7 +51,7 @@ const AGenericLoader = forwardRef(
       }
     }
 
-    switch (childPlacement) {
+    switch (placement) {
       case "top":
         className += ` ${baseClass}--placement-top`;
         break;
@@ -74,17 +74,17 @@ const AGenericLoader = forwardRef(
 
     return (
       <div ref={ref} className={className} {...rest}>
-        {["top", "left"].includes(childPlacement) && children}
+        {["bottom", "right"].includes(placement) && children}
         <div className={containerClassName}>
           <TagName {...indicatorProps} size={size}></TagName>
         </div>
-        {["bottom", "right"].includes(childPlacement) && children}
+        {["top", "left"].includes(placement) && children}
       </div>
     );
   }
 );
 
-AGenericLoader.propTypes = {
+ALoader.propTypes = {
   /**
    * Indicator type to use in the loader. Can be ADotLoader, ASpinLoader, or AProgressbar.
    */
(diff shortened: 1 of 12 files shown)


brennarvo:
Haven't looked at the code yet, but first thoughts:

1. Can we name the component `ALoader`? I think that indicates that it is generic versus the more specific `ADotLoader`, `ASpinLoader` etc. variants
2. Instead of `component`, can we name it `variant`?

So ultimately:

```html
<ALoader variant="spinner">I spin</ALoader>

<ALoader variant="dot">I am sort of like wavy OSRS text</ALoader>
```

And also, maybe this is more confusing, but what about having a `placement` prop that indicates where the *loader* goes instead of the child?

So

```html
<ALoader variant="dot" placement="left">Text rendered to the right of the loader</ALoader>
```

would look like

loader-placement

This also aligns with how we have the [`placement` prop for `AMenu`](https://magna-react.vercel.app/components/menu#menu-props)

rwharrisjr:
`ALoader` makes sense. Initially I chose `childPlacement` since it the placement only changes if children are passed, which is not required. But `placement` does keep consistency with other component. PR updated.

framework/components/ALoader/ALoader.js
```diff
@@ -0,0 +1,111 @@
...
+      placement = "top",
+      children,
+      ...rest
+    },
+    ref
+  ) => {
+    let className = baseClass,
+      containerClassName = `${baseClass}__container`;
+
+    const TagName = typeof variant === "string" ? INDICATORS[variant] : variant,
```
brennarvo, line 33: Ah, I probably did comment too early before reading code.

This conditional allows you to pass a custom component for the loader, though I'm sure the intent is that it will only ever be `ASpinLoader`, `AProgressbar`, or `ADotLoader`, in which case `variant` still sounds good. But I do see why it was `component`. Sorry for the early assumption there.

rwharrisjr: I think variant is good, it's more consistent with other prop names. This just allows for a little extra customization if the dev needs it.

mvelkCisco:
@brennarvo @rwharrisjr I'm a bit late for the party, but I think in this case it might have been better to only have separate components. `ALoader` should not exist. Skeleton, spinner and progress bar are completely distinct loading patterns that should be used for different use cases and they look completely different. See: https://magnetic.cisco.com/0a43ab5cd/p/43b10f-progress-and-loading-indicators/b/914c27

Also the "dot" loader could have been omitted as a variant. There's no such a thing in magnetic and it's usage should be discouraged. Backwards compatibility is not needed for a newly added component, it does not break anything. But it may bee too late to change it now.

I've added https://github.com/cisco-sbg-ui/magna-react/issues/332 to follow up.

PR #317, opened 2023-03-22
Refactor/drawer mounting

brennarvo:
This is a draft PR to explore options in conditionally rendering both `AModal` and `ADrawer`. The intent is that we don't always need to keep their children rendered in order to perform smooth transitions.

This will also give `AModal` a similar animation found in the Magnetic React team's modal.

### DIFF
--- a/framework/components/ADrawer/ADrawer.scss
+++ b/framework/components/ADrawer/ADrawer.scss
@@ -1,5 +1,14 @@
 @import "../../styles";
 
+$transition-duration: 0.3s;
+$transition-timing: ease-in-out;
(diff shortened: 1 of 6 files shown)


framework/components/ADrawer/ADrawer.scss
```diff
@@ -113,33 +113,112 @@
...
+}
+
+@keyframes slideOutToBottom {
+  0% {
+    transform: translateY(0);
+  }
+  100% {
+    transform: translateY(100%);
+  }
+}
```
brennarvo, line 224: Obvious refactoring I can do here with SCSS. This was just to get it all working.

PR #319, opened 2023-03-23
feat: conditionally render modal and drawer children

brennarvo:
### ⚠️ This introduces breaking changes for both `AModal` and `ADrawer`

~~I would feel better merging this after #321.~~ Done

[Demo](https://magna-react-git-refactor-drawer-mounting-transitions.securex-preview.app/components/modal#usage)

### `AModal`: New Props

There are two new props to make `AModal` more flexible for integrating applications.

*  `withCenteredContent`: Determines if the content rendered in the modal should automatically be centered vertically and horizontally.
* `withTransitions`: Determines if the modal will open and close with the fade-in and grow transitions

### `AModal`: Prop Changes
Props are changing to keep naming consistency with the other `AModal` props, making it easier to remember what can be passed.

* `trapFocus` renamed to `withFocusTrap`
* `lockScroll` renamed to `withScrollLock`

### `AModal`:  UX Changes

* CSS transitions for both the open and close states

### `AModal`: Breaking Changes

<details>
<summary>:warning: Children of <code>AModal</code> will mount and remount each time <code>isOpen</code> is toggled</summary>
<br />
Each time the modal is opened and closed, the children passed to `AModal` will always mount and unmount. This means that the state within this component will always reinitialize, as well as reinitiate React's lifecycle, resulting in extra calls to hooks like `useEffect`, `useLayoutEffect`, etc.
</details>

<details>
<summary>:warning: Prop renaming</summary>
<br />

* `trapFocus`  -> `withFocusTrap`
* `lockScroll` ->  `withScrollLock`
</details>

## `ADrawer` Updates

[Demo](https://magna-react-git-refactor-drawer-mounting-transitions.securex-preview.app/components/drawer#usage)

### `ADrawer`: New Props

One new prop to toggle CSS transitions

* `withTransitions`: Determines if the drawer will open and close with the slide-in and slide-out transitions

### `ADrawer`: Breaking Changes

Because `ADrawer` is composed of `AModal`, it inherits the same breaking changes. For clarity:

<details>
<summary>:warning: Children of <code>ADrawer</code> will mount and remount each time <code>isOpen</code> is toggled</summary>
<br />
Each time the drawer is opened and closed, the children passed to `ADrawer` will always mount and unmount. This means that the state within this component will always reinitialize, as well as reinitiate React's lifecycle, resulting in extra calls to hooks like `useEffect`, `useLayoutEffect`, etc.
</details>

<details>
<summary>:warning: Modal props passed to drawers will need to align to <code>AModal</code> prop changes</summary>
<br />
If `ADrawer` is being rendered as a modal (which it is by default), then any props previously being passed to `ADrawer` to toggle modal behavior will need to be updated to match the prop name changes mentioned in the `AModal` breaking changes:

* `trapFocus`  -> `withFocusTrap`
* `lockScroll` ->  `withScrollLock`
</details>

### DIFF
--- a/framework/components/AModal/AModal.js
+++ b/framework/components/AModal/AModal.js
@@ -14,6 +14,7 @@ import {
 } from "../../utils/helpers";
 
 import "./AModal.scss";
+import {useDelayUnmount} from "../../utils/hooks";
 
 /**
  * Reasons for not stopping propagation for the following keys:
@@ -54,23 +55,30 @@ const AModal = forwardRef(
     const hasPortalNode = appendTo || appRef.current;
     const isOpen = !!hasPortalNode && propsIsOpen;
     const _ref = useRef();
+
+    const shouldRender = useDelayUnmount(isOpen, 300);
+
     useFocusTrap({
       rootRef: _ref,
-      isEnabled: trapFocus && isOpen
+      isEnabled: trapFocus && shouldRender
     });
+
     useEffect(() => {
-      isOpen && lockScroll ? preventBodyScroll() : allowBodyScroll();
+      shouldRender && lockScroll ? preventBodyScroll() : allowBodyScroll();
       return allowBodyScroll;
-    }, [lockScroll, isOpen]);
+    }, [lockScroll, shouldRender]);
+
+    const animationClassName = isOpen
+      ? "animation-fade-in"
+      : "animation-fade-out";
+
     let visibilityClass = "";
-    if (!isOpen) {
-      visibilityClass = "a-modal--hidden";
+
+    if (!shouldRender) {
+      visibilityClass += "a-modal--hidden";
     }
 
     let contentClassName = `a-modal-container ${visibilityClass}`;
-    if (propsClassName) {
-      contentClassName += ` ${propsClassName}`;
-    }
 
     if (small) {
       contentClassName += " a-modal-container--small";
@@ -80,6 +88,10 @@ const AModal = forwardRef(
       contentClassName += " a-modal-container--large";
     } else if (xlarge) {
       contentClassName += " a-modal-container--xlarge";
+    }
+
+    if (propsClassName) {
+      contentClassName += ` ${propsClassName}`;
     }
 
     if (!appendTo && !appRef.current) {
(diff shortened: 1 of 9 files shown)


framework/components/AModal/AModal.js
```diff
@@ -35,16 +36,19 @@ const AModal = forwardRef(
       appendTo = null,
       as = "div",
       className: propsClassName,
+      delayUnmount = 200,
```
brennarvo, line 39: I don't like how unintuitive this prop feels. Unless you are me, and worked on conditional rendering children while preserving CSS transitions, you won't have a clue what this is for.

The only reason I needed it as a prop was so that I could also have CSS transition effects in `ADrawer`, which also renders `AModal`.

Probably worth documenting this, come to think of it.

mvelkCisco: Ah, interesting. Or maybe you can unmount the whole `AModal` in `ADrawer` instead?

mvelkCisco: Or maybe extract some core functionality from `AModal` to a shared `AModalBase` (or something like that) and use it in both. The coupling between `AModal` and `ADrawer` looks a bit unfortunate.

mvelkCisco:
Btw another thing with drawers I've been facing is the requirement that only 1 drawer should be open at a time. When a new drawer is opening, the previous one needs to close first, the animation (and the unmount) needs to finish, and then the new one needs to mount and start opening. So there's quite a bit of synchronisation going on (see https://github.com/advthreat/intelligence/pull/79). I'll need to keep that (and know the animation durations) anyway, even when this is merged.

I'm not sure if that can be done in a good way without introducing some kind of a context/provider and might be an overkill to add to magna react itself... just some food for though for the future.

mvelkCisco:
It looks like the children unmounting is working correctly. The timing seems correct. The containers remain mounted at all times:

image

In situations when there are hundreds of separate drawers in DOM it's a bit unfortunate still.

I understand there's a problem with the animation when you want to mount/unmount the whole drawer. To mitigate that, you would need to force react render the drawer in 2 passes, first closed, then with the "show" class. I do that with `setTimeout` with timeout 0 here: https://github.com/advthreat/intelligence/blob/master/src/contexts/DrawersContext.js#L30. It is kind of hacky, but it works.

Another option is to use semi-official ReactTransitionGroup for the animation: https://legacy.reactjs.org/docs/animation.html. I had this working for a while but then I decided it's not worth the dependency. It would be more resilient though.

Sorry I don't have enough time to look into this too deeply right now and I'm on PTO tomorrow.

PR #322, opened 2023-03-24
Added AListItemGroup to be able to group list items into unified set #318

bredy:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows semantic commit message guidelines
- [x] The changes are documented in component docs and changelog
- [x] The ESLint plugin has been updated if a new component is added
- [x] Test have been added or modified, if appropriate
- [x] Has been verified locally
- [x] The component should accept a class name as a prop, if appropriate
- [x] The component should accept a forward ref, if appropriate

**What kind of change does this PR introduce?**

This PR introduces new feature to group list items (`AListItem`) under one unifying title as requested in Magnetic design. 

 Closes #318 

**What is the current behavior?** 

Current behaviour doesn't support such grouping. 

**What is the new behavior (if this is a feature change)?**

New behaviour supports grouping using `AListItemGroup` component. Docs was updated and contains this new component usage for "Lists" as well as "Menus".

**Does this PR introduce a breaking change?** 

No, this change should be backward compatible.

PR #324, opened 2023-03-24
change divider width

### LINKED ISSUE #323: ADivider is 2px but probably should be 1px
We came across this in https://github.com/cisco-sbg-ui/magna-react/pull/322#discussion_r1147479570

The magna-react docs show `ADivider` used in the menu, so that's what products are doing (eg incident-manager). However based on figma the divider in menus should be 1px not 2px (incident-manager adds a global override to make all divides 1px).

I don't think a divider exists in magnetic, or I could not find it at least. All the separators I've seen on magnetic are 1px.  This doesn't look right:

image

I would propose to just change it in magna-react to 1px.

mvelkCisco:
fixes #323

### DIFF
--- a/framework/components/ADivider/ADivider.scss
+++ b/framework/components/ADivider/ADivider.scss
@@ -14,7 +14,7 @@
 
 .a-divider {
   position: relative;
-  border-width: $border-width 0 0 0;
+  border-width: 1px 0 0 0;
   border-style: solid;
   margin: 20px 0;
 }


rwharrisjr:
Since `ADivider` is used in a lot of places, what about adding a `thin` variant, with a class `.a-divider--thin` with `border-width: thin 0 0 0;`

`1px` for borders is sometimes problematic with browser scaling, but `thin` gets around the issues. That may have just been in older browser versions though as I'm not able to reproduce that behavior on `ADivider`, at least in Chrome.

mvelkCisco:
Adding `AListItemDivider` (name tweaked as it's used inside `AList`, not between lists) in https://github.com/cisco-sbg-ui/magna-react/pull/322

PR #327, opened 2023-03-27
AInlineTextInput and AInlineTextarea

rwharrisjr:


framework/components/AInlineInput/AInlineInput.js
```diff
@@ -0,0 +1,215 @@
...
+                small={small}
+                medium={medium}
+                large={large}
+                autoFocus // eslint-disable-line
+                {...inputComponentProps}
+              />
+            </AForm>
+          </div>
+        )}
+        {!isEditing && (
```
brennarvo, line 139: Any objections to the code style below? (simplified for example's sake):

```js
{isEditing ? <InputWhileEditing /> : <InlineText />}
```

framework/components/AInlineInput/AInlineInput.js
```diff
@@ -0,0 +1,215 @@
...
+                  e.stopPropagation();
+
+                  const {value: eventValue} = e.target;
+                  setEditingValue(eventValue);
+                }}
+                onKeyDown={(e) => {
+                  if (e.key === "Enter") {
+                    handleChange(e);
+                  }
+                }}
```
brennarvo, line 128: Same comment as earlier about calling prop handlers. In this case, checking if `inputComponentProps` passed into `AInlineInput` have any `onKeyDown`, `onChange`, or `onBlur` props, else we'll override them and people might get confused why theirs aren't being called

framework/components/AInlineInput/AInlineInput.js
rwharrisjr: That makes a the ternary a bit long for my liking, but I will split it into an if/else to make things more readable

framework/components/AInlineInput/AInlineInput.js
```diff
@@ -0,0 +1,215 @@
...
+        onClick={(e) => {
+          e.stopPropagation();
+
+          if (disabled) {
+            return;
+          }
+
+          !isEditing && setIsEditing(true);
+        }}
+        onKeyDown={(e) => {
```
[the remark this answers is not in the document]
rwharrisjr: Done

framework/components/AInlineInputBase/AInlineInputBase.js
```diff
@@ -0,0 +1,238 @@
...
+
+    if (small) {
+      className += ` ${baseClass}--size-small`;
+    } else if (large) {
+      className += ` ${baseClass}--size-large`;
+    } else {
+      className += ` ${baseClass}--size-medium`;
+    }
+
+    const handleChange = (e) => {
```
brennarvo, line 72: I'm rereading this with a fresh set of eyes this morning, and I'm starting to think that `handleChange` and the `onChange` prop passed to this component behave differently than I would expect.

When I pass on `onChange` to a component associated with an input field, I typically expect it to be called when the user types something. In this case, though, `handleChange` acts more like a "save".

```js
const [value, setValue] = useState('')

return (
  /* Here, `onChange` is not actually called until the user saves the edit */
  <AInlineTextInput onChange={(e) => setValue(e.target.value)} />
)
```

I think it would be a bit more clear if we named this `handleEditComplete` or something alike to indicate that we call this when the user has finalized the text. We can then remap the `onChange` prop to something similar:

```js
const [value, setValue] = useState('')

return (
  /* More apparent `onEditComplete` is called when the user is done typing */
  <AInlineTextInput onEditComplete={(e) => setValue(e.target.value)} />
)
```

framework/components/AInlineInputBase/AInlineInputBase.js
```diff
@@ -0,0 +1,238 @@
...
+const baseClass = "a-inline-text-input";
+
+const AInlineInputBase = forwardRef(
+  (
+    {
+      inputComponent = ATextInput,
+      inputComponentProps = {},
+      clearable,
+      disabled,
+      value = "",
```
brennarvo, line 20: Kind of along the same lines as `onChange`, this makes it seem that passing `value` will override what the user is typing. I know the example below isn't realistic, but it illustrates my point:

```js
const [value, setValue] = useState("");

return (
  <>
    <input
      id='raw-input'
      type="text"
      value={value}
      onChange={(e) => setValue(e.target.value)}
    />
    <AInlineTextInput value={value} />
  </>
);
```
When I type into `#raw-input`, then `<AInlineTextInput />` properly updates to align with what the user is typing into `#raw-input`.

However, there will be a point in time where the `value` passed to `<AInlineTextInput />` is not in sync, specifically when the user is editing the value inside `<AInlineTextInput />`:

example

In the screenshot above, the `value` being passed to `<AInlineTextInput />` is (rightfully) overridden by the internal state stored in `<AInlineTextInput />`.

Do you find this a little confusing, or is this a me thing? I'm thinking maybe naming `value` `defaultValue` would make it more obvious as to what is happening? `defaultValue` usually indicates that the internal state mechanism is not controlled externally, which is technically what is happening here.

PR #328, opened 2023-03-28
Add ACard component

### LINKED ISSUE #288: APanel inconsistencies with magnetic
Magnetic provides 2 types of "cards", "basic" and "container": https://magnetic.cisco.com/0a43ab5cd/p/60f4a1-card/b/97571f

These differ not just in border but also in padding, border radius, etc.

Magna-react has `APanel` with `borderless` prop with default `null`.

Looking through the apps, most are using the bordered version (default) for almost everything, but should be using "container" type for most of the cards that define the page layout: https://magnetic.cisco.com/0a43ab5cd/p/60f4a1-card/b/918302

Honestly I'm not sure how to out of this at this point. We could have had a `type=basic|container` prop with `container` default, but the `type` prop is taken by `Enum("default", "grey", "white", "dialog")`. 

We could change the defaults to resemble "container" more and add some prop to switch it to "basic" mode, and let the apps deal with the changes... I really don't know.

Thoughts?

rrubas:
Resolves #288

### DIFF
--- a/framework/components/ACard/ACardBasic.js
+++ b/framework/components/ACard/ACardBasic.js
@@ -64,6 +64,12 @@ ACardBasic.propTypes = {
    * Sets the base component.
    */
   component: PropTypes.elementType,
+
+  /**
+   * Sets the colour state on the card.
+   * It reflects the colors of the magnetic design states "dormant", "positive", "warning", "negative" and "informational".
+   * Value "custom" sets the state color from "stateCustomColor"
+   */
   state: PropTypes.oneOf([
     "dormant",
     "positive",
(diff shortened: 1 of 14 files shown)


rrubas:
Snímek obrazovky 2023-04-01 v 10 54 24
Snímek obrazovky 2023-04-01 v 10 54 59

framework/components/ACard/ACardBasic.js
```diff
@@ -0,0 +1,95 @@
...
+);
+
+ACardBasic.propTypes = {
+  /**
+   * Sets the base component.
+   */
+  component: PropTypes.elementType,
+
+  /**
+   * Sets the colour state on the card.
```
mvelkCisco, line 67: "colour" -> "color", or maybe better just: "Sets the card state color."

mvelkCisco (approved):
Looks good to me. 

@rwharrisjr what do you think?

rwharrisjr (approved):
looks great!

PR #339, opened 2023-04-04
feat(AMount): allow children to determine edge detection menu calculations

brennarvo:
This PR allows a child of `AMount` to determine the edge detection settings by passing `withNewWrappingContext={true}` to the component.

Example:

```jsx
<AMount>
    <div> {/* <-- this div is the new edge determinator */}
        Is this PR a feature or fix? Probably a feafix or fixeat.
    </div>
</AMount>
```

This is better demonstrated on the [example docs](https://magna-react-pfy3lvsn7.securex-preview.app/components/mount).

### DIFF
--- a/docs/mock_modules.js
+++ b/docs/mock_modules.js
@@ -1,17 +1,18 @@
+import {useEffect, useState} from "react";
 /**
  * These are modules that can be used inside of documentation
  * to provide more enriching examples. The purpose of keeping
  * these functions in this file is to abstract away unnecessary
  * implementation details for setting up an example in an effort
  * to focus more on the component's usage itself.
- * 
+ *
  * @example using the randomColors in documentation
  * const randomColors = mockImport('randomColors');
  * randomColors.get().then(randomColors => // use random colors);
  */
 const modules = {
-  randomColors: (function() {
-    const getHex = (num) => num.toString(16).padStart(2, '0');
+  randomColors: (function () {
+    const getHex = (num) => num.toString(16).padStart(2, "0");
     const getRandomRgb = () => {
       const r = Math.floor(Math.random() * 256);
       const g = Math.floor(Math.random() * 256);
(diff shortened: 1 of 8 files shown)


docs/mock_modules.js
brennarvo, line 1: This is really unimportant for the PR. It just allowed me to add the `useDraggable` component to a docs playground without having to define it in the same code editor snippet that readers of the docs site would see.

framework/components/AMenuBase/AMenuBase.js
```diff
@@ -116,17 +117,30 @@ const calculateMenuPosition = (
     window.getComputedStyle(combinedRef.current).marginTop
   );
 
-  const xOffset = appRef.current.offsetParent?.isSameNode(document.body)
-    ? window.pageXOffset
-    : wrapCoords.left - appCoords.scrollLeft;
-  const yOffset = appRef.current.offsetParent?.isSameNode(document.body)
-    ? window.pageYOffset
-    : wrapCoords.top - appCoords.scrollTop;
```
brennarvo, line 124: ` appCoords.scrollTop` was the culprit it the original wonkiness. That isn't defined on a `DOMRect` object, hence the miscalculations that have happened before.

framework/components/AMount/AMount.js
```diff
@@ -5,13 +5,16 @@ import AAppContext from "../AApp/AAppContext";
...
+import "./AMount.scss";
+
 const AMount = forwardRef(
   (
     {
       children,
       className: propsClassName,
       component,
       wrapClassName: propsWrapClassName,
+      withNewWrappingContext = false,
```
brennarvo, line 17: Here is what makes this PR "opt-in". It'll be `false` by default.

pages/[[...route]]/index.js
Comment (brennarvo) [inline pages/[[...route]]/index.js:1]: I also went ahead and wrapped the docs site with another `AMount` so that the menus from examples don't bleed into the docs site sidebar.

This is more of a test to see if it'll actually play out correctly in a MFE, since they'll be rendering `<AApp />` very similarly.

rwharrisjr (approved):

Issue #336, opened 2023-04-03
`useAToaster` default timeout


vdineva:
Please consider supporting default timeout for `useAToaster` of 8,000ms for warning toasts and 6,000ms for information toasts

This is the relevant discussion in from the `Magnetic Design System Support` channel from 01.03.23:

<img width="906" alt="222283495-1aa5d20c-02d5-4257-84a6-57684d21dccd" src="https://user-images.githubusercontent.com/791418/229589028-e1fc8a5e-6c8b-4820-99c0-26f1a60b6c65.png">

PR #342, opened 2023-04-06
feat: Pause toast timeout on hover

rwharrisjr:
Based on the pabst-components Notifier functionality, this will pause the dismiss timeout of an `AToast` generated by `useAToaster` when the user hovers over the toast message. 

Originally this was a request from users and UX to give more time to read some notifications.

Functionality:

1. Toast message is generated with a default 6 second timeout (#336)
2. User hovers over toast after 2 seconds
3. Timeout is paused indefinitely until user hovers away
4. Timeout resumes with 4 seconds left
5. Toast dismisses after 4 seconds

### DIFF
--- a/framework/components/AToaster/customTimeout.js
+++ b/framework/components/AToaster/customTimeout.js
@@ -1,0 +1,43 @@
+class PausableTimeout {
+  constructor(callback, delay) {
+    this.callback = callback;
+    this.delay = delay;
+    this.remaining = delay;
+
+    this.start();
+  }
+
+  start() {
+    this.running = true;
+    this.started = new Date();
+    this.id = setTimeout(() => {
+      this.callback();
+    }, this.remaining);
+  }
+
+  pause() {
+    this.running = false;
+    clearTimeout(this.id);
+    this.remaining -= new Date() - this.started;
+  }
+
+  clear() {
+    this.running = false;
+    clearTimeout(this.id);
+  }
+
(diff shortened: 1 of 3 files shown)


framework/components/AToaster/customTimeout.js
```diff
@@ -0,0 +1,43 @@
+class PausableTimeout {
+  constructor(callback, delay) {
+    this.callback = callback;
+    this.delay = delay;
```
brennarvo, line 4: Is this property here for somebody instantiating the class? Only asking because I notice it isn't being used, but I can see this being useful:

```js
const pausableTimeout = new PausableTimeout(() => alert("done"), 2000);

// Somewhere else
console.log(pausableTimeout.delay); // 2000
```

framework/components/AToaster/customTimeout.js
```diff
@@ -0,0 +1,43 @@
...
+    this.remaining = delay;
+
+    this.start();
+  }
+
+  start() {
```
brennarvo, line 10: Also should reset `this.remaining` to be equal to the initial `delay` so that the `pause` method is aware a new interval has started when it updates `this.remaining` on line 21

brennarvo: Sorry, I take this back. I see how that would mess things up. It should be in the `clear` method.

brennarvo (approved):

PR #347, opened 2023-04-11
Use color-scheme for scrollbars

rwharrisjr:
@trent-boyd suggested using color-scheme for automatic scrollbar styling


trent-boyd (approved):
Yep, that's pretty much it 👍

PR #351, opened 2023-04-17
Fix hover color state for dot icons inside AButton

### LINKED ISSUE #349: Dots-three icon no style on hover
`tertiaryAlt` needs a hover color for the the `dots-three` icon
Robert note: `a-button on hover is just changing "color", and the icon is picking up currentColor for stroke`
icons

rwharrisjr:
Resolves #349


androche (approved):
Tested by copying css into icon I was working with

PR #359, opened 2023-04-25
fix: button spinner color

androche:
Closes #359

Quick fix on button loader colors:

Screenshot 2023-04-25 at 2 21 47 PM
Screenshot 2023-04-25 at 2 21 39 PM

Note* Thinking since it looks like the button loaders are standardized in magnetic, magna could probably handle it all through a loading prop.

### DIFF
--- a/framework/components/AButton/AButton.scss
+++ b/framework/components/AButton/AButton.scss
@@ -444,6 +444,13 @@ $btn-transition: color $transition-duration--extra-fast
     stroke: currentColor;
   }
 
+  .a-spinner {
+    margin-right: 2px;
+    circle {
+      stroke: currentColor !important;
+    }
+  }
+
   .a-icon.a-icon--dots-three,
   .a-icon.a-icon--dots-six,
   .a-icon.a-icon--dots-nine {


rwharrisjr (approved):

framework/components/AButton/AButton.scss
```diff
@@ -444,6 +444,13 @@ $btn-transition: color $transition-duration--extra-fast
     stroke: currentColor;
   }
 
+  .a-spinner {
+    margin-right: 2px;
```
vdineva, line 448: Thank you for doing this work @androche !

I think this margin should actually be 5px:
Screenshot 2023-04-26 at 8 49 35
Reference: https://www.figma.com/file/yU3qYkj6T2kfN4XYtUkEeD/Dark-Mode-%5BWiP%5D?node-id=33305-358611&t=wMfnk68xQIygesaq-4

androche: Thanks so much @vdineva - I didn't have this figma link!!   So I did try the 5px initially, but it looked like a bit too much space compared to the magnetic implementation.  Double checked and sure enough the width on the spinner_figure is a little wider than the icon itself so that gives it a couple extra pixels.

**5px**
Screenshot 2023-04-26 at 9 30 18 AM

**Magnetic**
Screenshot 2023-04-26 at 9 15 34 AM

Agree though that 2px is definitely too little width -looks like 3px would be the ideal spacing:
**3px**
Screenshot 2023-04-26 at 9 32 51 AM

PR #369, opened 2023-05-08
feat: add tooltip prop for AInlineInputBase

rwharrisjr:
Adds a prop `showTooltip` to add a tooltip of the display value to the value when the inline edit is not active


androche (approved):
Pulled and tested

Screenshot 2023-05-08 at 6 37 31 PM

PR #373, opened 2023-05-10
style: handle long tags in ATag component according to magnetic design

### LINKED ISSUE #372: Update `ATag` component to handle long tags
**Describe the bug**

We have reported issue https://github.com/advthreat/intelligence/issues/176 in the Intelligence app. The reported issue complains about a long tag being wrongly displayed, with the text extending outside of the tag box. However, this application only uses the `ATag` component and inserts long text into it. Therefore, the issue should be correctly fixed in magna-react.

**To Reproduce**

1. Take example file `ATag.mdx`
2. Add new or modify current tag ("User tag") - add very long text into it, e.g.:
```
      <ATag className="cisco-blue white--text ma-1">
        <AIcon left>user</AIcon>User Tag VERY VERY LONG TAG WHICH CONTAINS BREAKABLE SPACES SO IT CAN BE DIVIDED TO MULTIPLE LINES
        <AIcon
          right
          className="cursor-pointer"
          onClick={() => alert("Close icon clicked!")}>
          close
        </AIcon>
      </ATag>
```
3. Check `ATag` example page.

**Current behavior**

Tag text is rendered outside of the tag box (see screenshot). 

**Expected behavior**

Tag is rendered well according to "Tag wrapping" section in Magnetic design:  https://magnetic.cisco.com/0a43ab5cd/p/716124-tag/t/625d3c

**Screenshots**

Screenshot 2023-05-10 at 13 18 53

 - Device: Desktop
 - OS: macOS
 - Browser Chrome
 - Version 112.0.5615.49

**Additional context**

Also from "Tag wrapping" is not entirely clear how should be tags displayed - when I suppose tags are inline elements we can either let tag to be filled horizontally, e.g. it can look like this:

Screenshot 2023-05-10 at 13 32 14

or e.g. vertically with maximum width:
image

We should consult with UX how this should be updated.

bredy:
- [ /] The commit message follows semantic commit message guidelines
- [ /] The changes are documented in component docs and changelog
- [ /] The ESLint plugin has been updated if a new component is added
- [ /] Test have been added or modified, if appropriate
- [ /] Has been verified locally
- [ /] The component should accept a class name as a prop, if appropriate
- [ /] The component should accept a forward ref, if appropriate

Resolves #372


framework/components/ATag/ATag.scss
```diff
@@ -41,14 +40,15 @@ $tag-icon-font-size: $font-size--xxs;
...
-  height: $tag-height;
   font-size: $tag-font-size;
   line-height: 22px;
-  align-items: center;
+  align-items: flex-start;
   padding: $tag-padding;
   border-radius: $tag-border-radius;
   vertical-align: middle;
   cursor: default;
+  margin: 2px;
```
mvelkCisco, line 50: I would not add any external margin to components like this. It will cause troubles when used and will need an override in many places.

mvelkCisco (approved):
Jus to clarify, this only allows multiline tags per magnetic specification to fix https://github.com/advthreat/intelligence/issues/176. It doesn't address missing sizes etc.

Looks good to me.

rwharrisjr (approved):

PR #377, opened 2023-05-10
Clear icon to only show when input has text

rwharrisjr:


brennarvo:
My only current hesitation with merging this right now is that it will add a state update on every render, which might be the reason somebody is trying to avoid rendering `<ATextInput />` as a controlled input.

androche:
Good catch @brennarvo, setState is the only way i've come across it as well.  @rwharrisjr Looks like some ppl do useRef to avoid this although while testing medium and large inputs only rerender a few times in the beginning of typing so might be ok.  The location usage example has a setState in the mdx so that one is a different case and does rerender every keystroke.

Only feedback I have is it shouldn't detect whitespace:
Screenshot 2023-05-12 at 3 55 52 PM

androche (changes requested):
Good catch @brennarvo, setState is the only way i've come across it as well.  @rwharrisjr Looks like some ppl do useRef to avoid this although while testing medium and large inputs only rerender a few times in the beginning of typing so might be ok.  The location usage example has a setState in the mdx so that one is a different case and does rerender every keystroke.

Only feedback I have is it shouldn't detect whitespace:
Screenshot 2023-05-12 at 3 55 52 PM

rwharrisjr:
React doesn't update the state again if a primitive stays the same value, so this will only trigger a re-render on the first keystroke for uncontrolled components. I think that's acceptable to get the functionality matching UX

androche (approved):
In case this comes up-  Robert and I chatted offline and we searched for other examples of allowing whitespace to be cleared as seen here:

https://material.angular.io/components/input/examples

NeelSaini:
Verified in rel v1.123 
Do not have the permission to add "QA Verified " Label

PR #379, opened 2023-05-11
#367 - add loading prop to button; middle align vertically

### LINKED ISSUE #367: Button loading state shifts position of other elements
A common pattern in Magnetic is to display a primary button and a secondary/tertiary button next to each other.
According to their specs: https://magnetic.cisco.com/0a43ab5cd/p/56608b-button/t/330d12 the primary button should become disabled when the action is async and we need to display a loader inside it

I've tried the following 2 approaches but they both shift the position of the secondary/tertiary button adjacent to the primary button and the 2 buttons are no longer vertically aligned:
Screenshot 2023-05-08 at 11 50 39

```
() => {
  return (
    <>
      <AButton secondary className="mr-2">Cancel</AButton>
      <AButton>
        <ASpinner size="small" />
        Click Me
      </AButton>
    </>
  );
}
```

```
() => {
  return (
    <>
      <AButton tertiary className="mr-2">Cancel</AButton>
      <AButton>
        <ASpinner size="small">Click Me</ASpinner>
      </AButton>
    </>
  );
}
```

rwharrisjr:
Resolves #367

### DIFF
--- a/framework/components/AButton/AButton.js
+++ b/framework/components/AButton/AButton.js
@@ -2,6 +2,8 @@ import PropTypes from "prop-types";
 import React, {forwardRef, useContext} from "react";
 
 import AButtonGroupContext from "../AButtonGroup/AButtonGroupContext";
+import ASpinner from "../ASpinner";
+
 import "./AButton.scss";
 
 const AButton = forwardRef(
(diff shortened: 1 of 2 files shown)


androche (approved):
Pulled and tested

PR #384, opened 2023-05-12
feat: replace icon only button icon with spinner if loading

brennarvo:
This adds documentation for the loading button, as well as handles the case where a loading button maybe be rendered within an "icon only" button. In that case, the icon only button's icon gets entirely replaced by the loading spinner. This isn't in the Magnetic specs, but it sorta just "feels right". Let me know if otherwise; I can easily remove.

### DIFF
--- a/framework/components/AButton/AButton.js
+++ b/framework/components/AButton/AButton.js
@@ -115,7 +115,7 @@ const AButton = forwardRef(
     return (
       <TagName {...props}>
         {loading && <ASpinner size="small" />}
-        {children}
+        {(!icon || (icon && !loading)) && children}
       </TagName>
     );
   }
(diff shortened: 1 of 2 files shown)


androche:
@brennarvo another great point here.  Also a tough one!  Do we know what is typically done in the UX world?  Trying to put myself in a user perspective, I'm thinking I find it a "cool" experience when the icon is replaced, but that's all I got 😆.  

- If we do take this approach, one issue to look into is absolute the position because the width will expand when the loading does pop up in practice (gif below).


- After fixing the expand, then we might still have it looking a little funky since it will appear as though the "padding" is less on the left side. 


- There's also the issue with the icon only as it would be replaced (otherwise would look super funky with two icons haha)
Screenshot 2023-05-12 at 4 27 05 PM


 **Expanding**
loading

rwharrisjr (approved):

Issue #371, opened 2023-05-09
Create a `TooltipIfTruncated` component


vdineva:
Currently the admin app in the monorepo has a component `TooltipIfTruncated`: https://github.com/advthreat/securex-ui-monorepo/blob/main/apps/admin/src/components/TooltipIfTruncated.js

and we need the same functionality in Device Insights. Can we have this component be part of magna-react so each app can easily re-use it?

https://github.com/cisco-sbg-ui/magna-react/assets/791418/cabd62cf-2f32-49a3-81a5-a9cbda892504

vdineva:
* disclaimer: I'm not sure this existent `TooltipIfTruncated ` component is per Magnetic specs

vdineva:
it seems it is: 
https://magnetic.cisco.com/0a43ab5cd/p/89dccd-tab/t/13e38f
https://magnetic.cisco.com/0a43ab5cd/p/632c0b-breadcrumb/t/06bfaf

rwharrisjr:
Added functionality as a prop on `ATriggerTooltip`

androche:
@vdineva just hoping you could verify if https://github.com/cisco-sbg-ui/magna-react/pull/394 is what you were looking for?  If so we can get that merged in for the next build.

vdineva:
@androche  works like a charm. Thank you!

rwharrisjr:
Available in `1.0.0-beta.46`

PR #394, opened 2023-05-18
#371 - add onlyIfTrunctated to ATriggerTooltip

rwharrisjr:
Adds `onlyIfTrunctated` prop to ATriggerTooltip to only show the tooltip if the content is truncated in some way.

@vdineva I think this will solve your use case, but it might not be a robust enough check. We can tweak it going forward if needed.

### DIFF
--- a/framework/hooks/useToggle/useToggle.js
+++ b/framework/hooks/useToggle/useToggle.js
@@ -8,7 +8,6 @@ const useToggle = (openDelay = 400, closeDelay, canOpen) => {
     timeout.current && clearTimeout(timeout.current);
 
     timeout.current = setTimeout(() => {
-      console.log("trying to open?");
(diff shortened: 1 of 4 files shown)


framework/hooks/useToggle/useToggle.js
```diff
@@ -1,16 +1,18 @@
...
+const useToggle = (openDelay = 400, closeDelay, canOpen) => {
   const [isOpen, setIsOpen] = useState(false);
   const timeout = useRef();
 
   const open = useCallback(() => {
     timeout.current && clearTimeout(timeout.current);
 
     timeout.current = setTimeout(() => {
-      setIsOpen(true);
+      console.log("trying to open?");
```
vdineva, line 11: you probably want to remove this console.log

androche (approved):

PR #397, opened 2023-05-19
feat: allow to specify which element to focus when a modal is opened, or opt-out from the autofocus while keeping the trap available

### LINKED ISSUE #395: `ATextarea`'s autofocus does not work reliably inside `AModal` with focus trap enabled
**Describe the bug**
When `AModal` has the `withFocusTrap` enabled, the mechanisms tries to focus the first focusable element inside the modal content. This however competes with the `ATextarea` `autoFocus` prop and unfortunatelly, the focus trap wins.

**To Reproduce**
1. Have the following modal.
```js
export default function MyModal({isOpen}) {
  return (
    <AModal isOpen={isOpen} aria-labelledby="modalTitle" withFocusTrap>
      <APanel type="dialog">
        <APanelTitle id="modalTitle">TITLE</APanelTitle>
        <AButton tertiary>Focusable</AButton>
        <ATextarea autoFocus />
        <AButton primary>Also focusable</AButton>
      </APanel>
    </AModal>
  );
}
```
2. Open this modal.

**Actual behavior**
The textarea gets focused for a split of a second, but then the focus is stolen to the `AButton`, which is first focusable element in the modal.

**Expected behavior**
The element which is already focused in the modal (in our case the `ATextarea` via it's autofocus feature) stays focused.

**Screenshots**
Screenshot 2023-05-19 at 9 11 56

### LINKED ISSUE #396: `AModal`'s focus trap automatically focuses first focusable element regardless if it's semantically appropriate
**Describe the bug**
`AModal`'s focus trap chooses the first focusable element and focuses it. However this is not always the right element to focus in terms of the semantics of the content.

**To Reproduce**
1. Have the following modal.
```js
export default function MyModal({isOpen}) {
  return (
    <AModal isOpen={isOpen} aria-labelledby="modalTitle" withFocusTrap>
      <APanel type="dialog">
        <APanelTitle id="modalTitle">TITLE</APanelTitle>
        <APanelFooter>
          <AButton tertiary>Tertiary</AButton>
          <AButton primary>Primary</AButton>
        </APanelFooter>
      </APanel>
    </AModal>
  );
}
```
2. Open this modal.

**Actual behavior**
The tertiary button is focused.

**Expected behavior**
Either no element is focused, or I can specify/override the element to get focused (possibly the primary button in this case).

**Screenshots**
Screenshot 2023-05-19 at 9 18 49

MikkCZ:
fix #395, fix #396

For #395, this change allows to opt-out from the implicit auto focus, which is part of the focus trap.

For #396, this change stops focusing the first focusable element and focuses the whole dialog instead.

### DIFF
--- a/framework/components/AModal/AModal.js
+++ b/framework/components/AModal/AModal.js
@@ -46,7 +46,7 @@ const AModal = forwardRef(
       onClickOutside,
       withCenteredContent = true,
       withFocusTrap = true,
-      focusTrapAutoFocus = "first",
+      focusTrapAutoFocus = "root",
       withOverlay = true,
       withScrollLock = true,
       withTransitions = true,
@@ -302,9 +302,9 @@ You should provide either an \`aria-label\` or \`aria-labelledby\` prop to \`${c
   withFocusTrap: PropTypes.bool,
 
   /**
-   * Specifies what element to autofocus when the modal is opened. Allows to choose between first focusable element, the modal root or opt-opt from element autofocus.
+   * Specifies what element to autofocus when the modal is opened. Allows to choose between the modal root or to opt-opt from element autofocus.
    */
-  focusTrapAutoFocus: PropTypes.oneOf(["first", "root", "none"]),
+  focusTrapAutoFocus: PropTypes.oneOf(["root", "none"]),
 
   /**
    * Determines if the modal should render with an faded backdrop.
--- a/framework/components/AModal/AModal.js
+++ b/framework/components/AModal/AModal.js
@@ -46,6 +46,7 @@ const AModal = forwardRef(
       onClickOutside,
       withCenteredContent = true,
       withFocusTrap = true,
+      focusTrapAutoFocus = "first",
       withOverlay = true,
       withScrollLock = true,
       withTransitions = true,
@@ -67,7 +68,8 @@ const AModal = forwardRef(
 
     useFocusTrap({
       rootRef: _ref,
-      isEnabled: withFocusTrap && shouldRenderChildren
+      isEnabled: withFocusTrap && shouldRenderChildren,
+      autoFocus: focusTrapAutoFocus
     });
 
     useEffect(() => {
@@ -165,6 +167,7 @@ const AModal = forwardRef(
           <Component
             role="dialog"
             aria-modal="true"
+            tabIndex={-1}
             ref={handleMultipleRefs(_ref, ref)}
             className={contentClassName}
             {...rest}
@@ -180,6 +183,7 @@ const AModal = forwardRef(
       <Component
         role="dialog"
         aria-modal="true"
+        tabIndex={-1}
         className={`${contentClassName}`}
         ref={handleMultipleRefs(_ref, ref)}
         onKeyDown={(e) => {
@@ -298,6 +302,11 @@ You should provide either an \`aria-label\` or \`aria-labelledby\` prop to \`${c
   withFocusTrap: PropTypes.bool,
 
   /**
+   * Specifies what element to autofocus when the modal is opened. Allows to choose between first focusable element, the modal root or opt-opt from element autofocus.
+   */
+  focusTrapAutoFocus: PropTypes.oneOf(["first", "root", "none"]),
+
+  /**
    * Determines if the modal should render with an faded backdrop.
    */
   withOverlay: PropTypes.bool,
@@ -310,11 +319,6 @@ You should provide either an \`aria-label\` or \`aria-labelledby\` prop to \`${c
 
   /**
    * Determines if the modal should open and close with CSS transitions.
-   */
-  withTransitions: PropTypes.bool,
-
-  /**
-   * Determines if the modal should open and close with CSS animations.
    */
   withTransitions: PropTypes.bool
 };
--- a/framework/components/AModal/AModal.js
+++ b/framework/components/AModal/AModal.js
@@ -1,4 +1,4 @@
-import React, {forwardRef, useContext, useEffect, useRef} from "react";
+import React, {forwardRef, useContext, useEffect, useRef, useMemo} from "react";
 import ReactDOM from "react-dom";
 import PropTypes from "prop-types";
 
@@ -46,7 +46,7 @@ const AModal = forwardRef(
       onClickOutside,
       withCenteredContent = true,
       withFocusTrap = true,
-      focusTrapAutoFocus = "root",
+      autoFocusElementRef,
       withOverlay = true,
       withScrollLock = true,
       withTransitions = true,
@@ -66,10 +66,21 @@ const AModal = forwardRef(
       isEnabled: withTransitions || delayUnmount
     });
 
+    const elementRefToAutoFocus = useMemo(() => {
+      if (autoFocusElementRef) {
+        return autoFocusElementRef;
+      } else if (typeof autoFocusElementRef === "undefined") {
+        return _ref;
+      } else {
+        // if (autoFocusElementRef === null) - disable autofocus
+        return undefined;
+      }
+    }, [_ref, autoFocusElementRef]);
+
     useFocusTrap({
       rootRef: _ref,
       isEnabled: withFocusTrap && shouldRenderChildren,
-      autoFocus: focusTrapAutoFocus
+      autoFocusElementRef: elementRefToAutoFocus
     });
 
     useEffect(() => {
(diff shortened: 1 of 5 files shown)


MikkCZ:
I made this PR in two commits on purpose. In case the behavior of auto focusing the first focusable element is actually desired by some designs, so I can revert it back easily. Although I don't think it's the expected default behavior to focus the first focusable element.

framework/components/AModal/AModal.js
```diff
@@ -297,6 +300,11 @@ You should provide either an \`aria-label\` or \`aria-labelledby\` prop to \`${c
    */
   withFocusTrap: PropTypes.bool,
 
+  /**
+   * Specifies what element to autofocus when the modal is opened. Allows to choose between the modal root or to opt-opt from element autofocus.
+   */
+  focusTrapAutoFocus: PropTypes.oneOf(["root", "none"]),
```
MikkCZ, line 306: I made this prop an enum intentionally, although there are only two values now, so it's easier extendable to any more clever algorithm how to find the element to auto focus when the modal is open. With only a boolean, this would be much harder to change in the future in a backward compatible manner.

framework/components/AModal/AModal.js
```diff
@@ -311,11 +319,6 @@ You should provide either an \`aria-label\` or \`aria-labelledby\` prop to \`${c
   /**
    * Determines if the modal should open and close with CSS transitions.
    */
-  withTransitions: PropTypes.bool,
-
-  /**
-   * Determines if the modal should open and close with CSS animations.
-   */
   withTransitions: PropTypes.bool
```
MikkCZ, line 322: It was twice here.

framework/components/AModal/AModal.js
brennarvo: How about we have a prop called: `initialFocusElemenetRef` that indicates which element should be focused when the modal is first initially opened. Would look like:

```jsx
function MyModal() {
  const textAreaRef = useRef();
  
  return (
    <>
      <AButton onClick={() => setIsOpen(true)}>Open</AButton>

      <AModal
        isOpen={isOpen}
        aria-labelledby="modalTitle"
        initialFocusElementRef={textAreaRef}
      >
        <AButton onClick={() => setIsOpen(false)}>Close</AButton>

        <ATextarea ref={textAreaRef} />
      </AModal>
    </>
  );
}
```

We could then subsequently pass this element to the `useFocusTrap` hook (see other PR comment).

framework/hooks/useFocusTrap/useFocusTrap.js
```diff
@@ -32,38 +32,42 @@ function createTrap(walker) {
   };
 }
 
-const useFocusTrap = ({rootRef, isEnabled = true}) => {
+export default function useFocusTrap({rootRef, autoFocus, isEnabled = true}) {
```
brennarvo, line 35: Following-up on the `initialFocusElementRef` comment:

Here, we could rename `autoFocus` to `initialFocusElementRef`. Then, we could check for its existence when determining the focus trap's initial focus before defaulting to `rootRef`.

framework/hooks/useFocusTrap/useFocusTrap.js
```diff
@@ -32,38 +32,42 @@ function createTrap(walker) {
...
+              ? NodeFilter.FILTER_SKIP
+              : NodeFilter.FILTER_ACCEPT;
           }
-          return node.tabIndex < 0
-            ? NodeFilter.FILTER_SKIP
-            : NodeFilter.FILTER_ACCEPT;
         }
-      }
-    );
-    const firstFocusableNode = treeWalker.nextNode();
```
brennarvo, line 54: Follow-up from above, this could become:

```js
const firstFocusableNode = initialFocusElementRef?.current || treeWalker.nextNode();
```

brennarvo (commented):
This change makes a lot of sense. The [ARIA Authoring Practices Guide (APG) even makes note of this distinctino for the dialog pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/#keyboardinteraction):

> * If a dialog contains the final step in a process that is not easily reversible, such as deleting data or completing a financial transaction, it may be advisable to set focus on the least destructive action, especially if undoing the action is difficult or impossible. The [Alert Dialog Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/) is often employed in such circumstances.
>
> * If a dialog is limited to interactions that either provide additional information or continue processing, it may be advisable to set focus to the element that is likely to be most frequently used, such as an OK or Continue button.

I do have one suggestion on how we might go about determining which element to focus to.

framework/components/AModal/AModal.js
MikkCZ: That wouldn't actually fix the fact, that there's an implicit auto focus done by `AModal`, which collides with `<ATextarea autoFocus />`. Also it wouldn't be much flexible for any custom focus logic, while keeping the focus trap in place. Personally, I don't think there should be any auto focus happening just by enabling the focus trap - that should be concern of the modal's content, not the modal itself.

In my case I hit this issue with an component, that renders the `ATextarea` inside it. But at the same time the component is not always rendered inside `AModal`. Also the `ATextarea` is not always rendered initially.

MikkCZ: I mentioned above, having it a string enum is more flexible to extend the behaviour for future. Using booleans (just for "consistency") is not a good idea, becuase the complexity grows very fast with all boolean props combinations.

Also in #396 I mention focusing the first focusable element (by default) is a bad idea. I think the reason why it's imlemented there now is just to remove focuse from the original element on the page, thet open the modal, and get the focus inside the focus trap. Focusing the root div solves this better, becuase it does not assume the first focusable element is the right one. It usually isn't - if there is any "default" or preferred action in a modal, it would be the primary button, which mostly comes last in the dialog. But I don't want to assume the last element is the right one to focus by default either.

brennarvo: > I mentioned above, having it a string enum is more flexible to extend the behaviour for future.

Then I think the `initialFocusElementRef` (`autoFocusElementRef` might sound better) makes sense so that we don't have to create additional enum identifiers for every scenario.  Allowing the consumer to pass a React `ref` gives them full control, and still enables the focus trap behavior.

> Using booleans (just for "consistency") is not a good idea, becuase the complexity grows very fast with all boolean props combinations.

Then `initialFocusElementRef` can be passed `null` to specify that no element should initially be focused to by the component, but would also still enable the focus trap behavior. This is equivalent to `"none"` with your `focusTrapAutoFocus` prop.

> Also in #396 I mention focusing the first focusable element (by default) is a bad idea.

This is outlined in the ARIA Authoring Practice Guide (APG):

> 1. When a dialog opens, focus moves to an element contained in the dialog. **Generally, focus is initially set on the first focusable element.** However, the most appropriate focus placement will depend on the nature and size of the content. Examples include:

I think what you mean is having the first focusable element as one of the consumer's child components is a bad idea (but correct me if I'm wrong). If that's the case, then it might be better for us to add a `tabindex="-1"` to the root element rendered in `AModal`, and if `initialFocusElementRef` is `null`, we'll set focus to the `root` (since we are able to do so with a `-1` ). Using `-1` as a value also means it wouldn't be tabbable in the tab sequence, which is good, since the only reason we altered the `tabindex` of the root element was to ensure we set focus within the modal content.

----

To summarize, here are some examples for how this would work.

1. This sets the initial focus to the root (i.e., `<div tabindex="-1" />`) rendered by `AModal`:

    ```js
    <AModal>
      <ATextarea />
    </AModal>
    ```
    Here, the `autoFocusElementRef` is omitted (`undefined`), meaning `AModal` will use its default behavior.

2. This tells `AModal` to skip any auto focus behavior, which would result in the `autoFocus` prop on `<ATextarea />` to be respected:

    ```js
    <AModal autoFocusElementRef={null}>
      <ATextarea autoFocus>
    </AModal>
    ```
3. This tells `AModal` to initially set focus to our `ATextarea` child (note that we don't pass `autoFocus` to the `<ATextarea />`. This is equivalent to the example 2.

    ```js
    const textareaRef = useRef();

    <AModal autoFocusElementRef={textareaRef}>
      <ATextarea ref={textareaRef}> {/* not passing `autoFocus` prop */}
    </AModal>
    ```

brennarvo (approved):
Thanks for taking time to update this 👍

MikkCZ:
@rwharrisjr please re-review.

PR #401, opened 2023-05-19
#400 - remove padding/margin of checkbox when in data table

### LINKED ISSUE #400: ADataTable has incorrect spacing when checkbox is used
**Describe the bug**

In our UI/UX sync today, we noticed that our table rows are not as short as the Magnetic design spec shows them to be.

I checked the magna docs and it looks like the table rows ARE using the current height. However, this is no longer the case when a checkbox is present, due to the checkbox component adding its own padding above and below it. It looks like this also might be the case for the `expandable` table variation. 

I also noticed that the table header for the checkbox has the text misaligned next to it.

This can all be seen on the magna docs site: https://magna-react.vercel.app/components/data-table

**To Reproduce**

1. Go to https://magna-react.vercel.app/components/data-table
2. Observe that all table examples show table rows the same height, except for the examples with checkboxes and accordion chevrons
3. Observe that the "with a checkbox" example also shows the checkbox header text misaligned with the other table header text

**Expected behavior**
Table rows should be the same height, regardless if a checkbox or accordion chevron is present.

**Screenshots**

Screenshot 2023-05-19 at 11 57 33 AM

<br />
Getting rid of this padding fixes the table row height problem when checkboxs are present:

Screenshot 2023-05-19 at 12 01 53 PM

<br />
Text misalignment issue with checkbox header:

Screenshot 2023-05-19 at 12 02 30 PM

rwharrisjr:
Resolves #400

### DIFF
--- a/framework/components/ADataTable/ADataTable.scss
+++ b/framework/components/ADataTable/ADataTable.scss
@@ -72,6 +72,22 @@ $hidden-table-col-width: 45px;
     }
   }
 
+  &__header,
+  &__cell {
+    .a-checkbox__wrap {
+      padding: 0 !important;
+    }
+
+    .a-checkbox__box {
+      margin-top: 0px !important;
+      align-self: center;
+    }
+
+    .a-checkbox__label {
+      top: 0px !important;
+    }
+  }
+
   .text-start {
     text-align: start;
   }


androche (approved):
Pulled and tested

PR #405, opened 2023-05-24
Add a resize observer to the tabs parent container to check if scroll buttons should show/hide

rwharrisjr:
When the parent container of ATabGroup changes size, the scrolling feature doesn't recalculate if the buttons should still be showing, or need to appear. This adds a ResizeObserver to check the parent container for size changes and trigger a show/hide of the buttons if needed.

### DIFF
--- a/framework/components/ATabs/ATabGroup.js
+++ b/framework/components/ATabs/ATabGroup.js
@@ -79,6 +79,12 @@ const ATabGroup = forwardRef(
       };
     }, [combinedRef, scrolling]);
 
+    useEffect(() => {
+      if (!showScrolling) {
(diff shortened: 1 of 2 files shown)


androche:
Very close! So after testing (i put debounce to zero) to avoid the visual jump.  It's better with zero so i'd recommend passing that instead of 100, but it still moves the content over to the left if we prevent that somehow.


scroll

rwharrisjr:
You're right, and debounce isn't needed when setting the primitive boolean to the useState. Removed.

androche (approved):
I think its great! There's a tiny jump when opening a collapsed panel but barely noticeable.  

approved

PR #408, opened 2023-05-24
fix: disable pointer events

### LINKED ISSUE #406: `AButton` with component prop still allows clicks when disabled
**Describe the bug**
When using `AButton` with a `Link` component prop the disabled state still allows clicks.

Example from DI:

```tsx
import { Link } from 'react-router-dom';

<AButton medium component={Link} disabled={true} to="/some/path" onClick={() => console.log('clicked')}>
  Click Me
</AButton>
```

**To Reproduce**
1. Create a `AButton` component using `Link` passed in as a component prop & an onClick callback
2. Disable the button using `disabled={true}` prop
3. Notice the button has disabled styling
4. Click the button - notice log is still getting printed when disabled.

**Expected behavior**
The button should not allow clicking on it when disabled.

**Screenshots**
Example from DI:

https://github.com/cisco-sbg-ui/magna-react/assets/95231401/3dc6b880-41ba-413d-bdf6-5a9940bd54fd

 - Device: Mac
 - OS: macOS
 - Browser: Safari
 - Version: `1.0.0-beta.41`

**Additional context**
Unfortunately unable to give code snip to test in the magna playground since the reproduction steps depend on a `Link` component from a routing package.

androche:
**Resolves #406 Prevent button events when disabled.**

**Problem**
`pointer-events: auto` allows click actions on links
`pointer-events: none` will prevent `cursor: not-allowed` from displaying

**Solution** 
Put a wrapper on the button to display `cursor: not-allowed` while preventing click action in child.

**Before**
alertTest

**After**
afterAlertTest

### DIFF
--- a/framework/components/AButton/AButton.js
+++ b/framework/components/AButton/AButton.js
@@ -34,7 +34,6 @@ const AButton = forwardRef(
     const {selectedValues, toggleValue} = useContext(AButtonGroupContext);
 
     let className = "a-button focus-box-shadow a-button--";
-    let disabledWrapperClass = "a-button--disabled";
 
     if (primary) {
       className += destructive ? "primary-destructive" : "primary";
@@ -113,18 +112,12 @@ const AButton = forwardRef(
       props.value = value;
     }
 
-    const btnComponent = (
+    return (
       <TagName {...props}>
         {loading && <ASpinner size="small" />}
         {(!icon || (icon && !loading)) && children}
       </TagName>
     );
-
-    if (disabled) {
-      return <span className={disabledWrapperClass}>{btnComponent}</span>;
-    }
-
-    return btnComponent;
   }
 );
 
--- a/framework/components/AButton/AButton.js
+++ b/framework/components/AButton/AButton.js
@@ -34,7 +34,7 @@ const AButton = forwardRef(
     const {selectedValues, toggleValue} = useContext(AButtonGroupContext);
 
     let className = "a-button focus-box-shadow a-button--";
-    let disabledWrapperClass = "";
+    let disabledWrapperClass = "a-button--disabled";
 
     if (primary) {
       className += destructive ? "primary-destructive" : "primary";
@@ -50,7 +50,6 @@ const AButton = forwardRef(
 
     if (disabled || loading) {
       className += " disabled";
-      disabledWrapperClass = "a-button--disabled";
     }
 
     if (small) {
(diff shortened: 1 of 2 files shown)


framework/components/AButton/AButton.js
```diff
@@ -112,12 +113,18 @@ const AButton = forwardRef(
...
 
-    return (
+    const btnComponent = (
       <TagName {...props}>
         {loading && <ASpinner size="small" />}
         {(!icon || (icon && !loading)) && children}
       </TagName>
     );
+
+    if (disabled) {
```
rwharrisjr, line 123: This adds more DOM, and I don't think we want to do that since it changes existing structure on pages.

framework/components/AButton/AButton.scss
```diff
@@ -477,10 +477,14 @@ $btn-transition: color $transition-duration--extra-fast
...
+  &--disabled {
+    //Wrapper class for disabled button to display `cursor: not allowed` when pointer events are disabled
+    cursor: not-allowed;
+    padding-bottom: 10px;
+  }
   &.disabled,
   &:disabled {
-    pointer-events: auto;
-    cursor: not-allowed;
+    pointer-events: none;
```
rwharrisjr, line 487: `pointer-events: none` is good, but not all variations get the attribute. Additionally nothing adds the `disabled` class in AButton it appears.

```js
() => {
  return (
    <div>
      <AButton data-testid="click-test-button" disabled component={() => <a href="https://www.cisco.com">Should not be able to click</a>} />
    </div>
  );
}
```

This scenario fails since <a> tags don't get the disabled attribute.

framework/components/AButton/AButton.js
rwharrisjr: I think it might be better to add `a-button--disabled` to the `className`, and then update the scss to use that everywhere there's a reference to "disabled".

rwharrisjr (changes requested):

framework/components/AButton/AButton.scss
rwharrisjr: ^ might also be a good example to add to the mdx

framework/components/AButton/AButton.js
androche: Yea I did that first, but trouble is `pointer events: none` and `cursor: not allowed` can't be in the same class to work.

androche: https://www.quora.com/How-can-I-set-pointer-events-to-none-as-well-as-cursor-to-not-allowed-simultaneously-in-CSS

androche:
Follow up on offline chat  - We are going with removing the disabled cursor for magna buttons altogether to avoid the wrapper complications and potential risk.

rwharrisjr (approved):

PR #412, opened 2023-05-31
fix: fallback to focusing root element when the provided reference is empty

MikkCZ:
regression from #397

### DIFF
--- a/framework/hooks/useFocusTrap/useFocusTrap.js
+++ b/framework/hooks/useFocusTrap/useFocusTrap.js
@@ -58,7 +58,9 @@ export default function useFocusTrap({
           .getAnimations({subtree: true})
           .map((animation) => animation.finished);
         Promise.allSettled(animationPromises).then(() =>
-          autoFocusElementRef.current?.focus({focusVisible: true})
+          (autoFocusElementRef.current ?? rootRef.current).focus({
+            focusVisible: true
+          })
         );
       }
       const trap = createTrap(treeWalker);
--- a/framework/hooks/useFocusTrap/useFocusTrap.js
+++ b/framework/hooks/useFocusTrap/useFocusTrap.js
@@ -58,7 +58,7 @@ export default function useFocusTrap({
           .getAnimations({subtree: true})
           .map((animation) => animation.finished);
         Promise.allSettled(animationPromises).then(() =>
-          autoFocusElementRef.current.focus({focusVisible: true})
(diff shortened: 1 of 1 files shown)


rwharrisjr (approved):

androche:
@MikkCZ  Just FYI we are unable to help verify this because cypress tests are already failing for us on incident manager on master for some reason.

MikkCZ:
Thanks for the notice Andrea. TBH I am unable to verify it locally either. The Cypress tests in incident-manager are only failing for me on Github Actions for the branch, but not locally.

androche:
I put a message out in chat for someone(anyone) to pull `master`, wipe their `node_modules`, reinstall, `npm run build`, then  `npm run test:ci`.  

This is what I get, and robert (doing the same steps) gets a different set of failures lol

Screenshot 2023-06-01 at 12 49 46 PM

framework/hooks/useFocusTrap/useFocusTrap.js
```diff
@@ -58,7 +58,9 @@ export default function useFocusTrap({
           .getAnimations({subtree: true})
           .map((animation) => animation.finished);
         Promise.allSettled(animationPromises).then(() =>
-          autoFocusElementRef.current.focus({focusVisible: true})
+          (autoFocusElementRef.current ?? rootRef.current).focus({
+            focusVisible: true
+          })
```
MikkCZ, line 63: I am thinking if the isssue might not actually coming from `rootRef` (respectively `_ref`) being already passed from `AModal` while it's value is still not set. @rwharrisjr you are probably more experienced with refs, so is that possible to happen? Should I add optional chaining to `?.focus`, or possibly replace refs in the list of dependencies by the `.current` values?

MikkCZ: `.current` of the `rootRef` (i.e. `_ref`) is accessed multiple times here. It would need to get nullish after all animations are settled. IMHO that should not happen, right?

rwharrisjr: I think optional chaining for `?.focus` would be a good safety just in case. @brennarvo is more familiar with focus trap with all the work he's done on the modals.

rwharrisjr:
The tests seem to be failing because of other changes in magna-react - behavior change was requested for the table sort icons, I suspect the external tests need to be updated to match.

I'm inclined to merge this since the code seems to work and shouldn't break anything.

PR #413, opened 2023-05-31
fix: hide icon when loading

### LINKED ISSUE #409: `AButton` with `loading` prop should show loader and hide icon in button
**Describe the bug**

In the `Loading state` section here: https://magnetic.cisco.com/0a43ab5cd/p/56608b-button
The specification says: "If there is an icon in the button, the loader replaces the icon."

**To Reproduce**
```
() => {
  const [loading, setLoading] = useState(false);
  return (
    <AButton onClick={() => setLoading(true)) loading={loading}>
        <AIcon left>edit</AIcon>
        Primary with Icon
      </AButton>
  );
}

```

**Expected behavior**
I would expect the spinner to replace the icon, i.e. the pen icon from the example code to be hidden

androche:
Closes #409 - Hide icon when loading. 

*Note I targeted magna icons only, but it works with generic svg as well if we think people wont be passing in magna icons. 

loading

### DIFF
--- a/framework/components/AButton/AButton.cy.js
+++ b/framework/components/AButton/AButton.cy.js
@@ -1,0 +1,29 @@
+import AButton from "./AButton";
+import AIcon from "../AIcon";
+
+describe("<AButton/>", () => {
+  it("should render", () => {
+    cy.mount(<AButton />);
+  });
+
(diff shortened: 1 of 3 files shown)


rwharrisjr (approved):
good fix

PR #417, opened 2023-06-06
fix tooltip positioning regression

mvelkCisco:
Should fix regression from https://github.com/cisco-sbg-ui/magna-react/pull/339.
Fixes https://github.com/advthreat/incident-manager/issues/948 (after magna-react bump).


rwharrisjr (approved):
ran in local and shell, resolves the placement issue

PR #419, opened 2023-06-06
feat: Add AMultiSelect component

### LINKED ISSUE #404: Multiselect component
We need a multiselect tag component in both Device Insights and User Insights 

similar to 
https://magnetic-react.cisco.com/?path=/story/core-select-multiselect--multi-select-with-tags

we can implement it ourselves but probably other XDR teams will need it as well

rwharrisjr:
Currently only supports tags. Resolves #404 

Screenshot 2023-06-06 at 4 41 45 PM

### DIFF
--- a/framework/components/AMultiSelect/AMultiSelect.js
+++ b/framework/components/AMultiSelect/AMultiSelect.js
@@ -409,8 +409,8 @@ const AMultiSelect = forwardRef(
           role="button"
         >
           {tags}
-          <input {...inputProps} />
         </div>
+        <input {...inputProps} />
         <AMenu ref={menuRef} {...menuComponentProps}>
           {prependContent}
           <div className="a-multiselect__menu-items__wrapper">
--- a/framework/components/AMultiSelect/AMultiSelect.js
+++ b/framework/components/AMultiSelect/AMultiSelect.js
@@ -9,10 +9,8 @@ import React, {
 
 import AInputBase from "../AInputBase";
 import {AFormContext} from "../AForm";
-import AButton from "../AButton";
 import AIcon from "../AIcon";
 import AMenu from "../AMenu";
-import ATag from "../ATag";
 import {AListItem} from "../AList";
 import AEmptyState from "../AEmptyState";
 import {useCombinedRefs} from "../../utils/hooks";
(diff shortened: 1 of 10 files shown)


androche (commented):
Works nicely!   Writing this as a comment because it doesn't have to be a blocker as i'm guilty of this myself in the other ticket, but the tabbing/arrows has some unwanted behavior.  I'll just post a couple gifs, but if you play around with it you can identify them:

Here I selected milk and cheese item, but it selected break and pasta item instead
oddTab-Arrows

Arrows/tabs skipping "Fruit".  (The beginning holdup is me trying to focus in on the form - had some trouble with that.)
tabSkipping

It's somehow possible to do both arrow and tabs at the same time - and they have different styles:
twoSelections

On blur toggle seems to be the same either way.
validationBlurSame

rwharrisjr:
@androche keyboard handling issues should be resolved now. Root cause was https://github.com/cisco-sbg-ui/magna-react/blob/canary/framework/components/AMenu/AMenu.js#L73-L75, when using DOMRect for placement - which is needed since the anchorRef can change size based on selection - it was set to ignore keyboard handling more menu items

framework/components/AMultiSelect/AMultiSelect.js
```diff
@@ -0,0 +1,575 @@
...
+    const filterFunction = propsFilterFunction
+      ? propsFilterFunction
+      : (val, item) => {
+          let displayValue;
+          if (typeof item === "string") {
+            displayValue = item;
+          } else if (typeof item === "object") {
+            displayValue = item[itemText];
+          }
+          //e.target.value.toLowerCase()
```
androche, line 83: can remove

framework/components/AMultiSelect/AMultiSelect.js
```diff
@@ -0,0 +1,575 @@
...
+            })}
+          <input {...inputProps} />
+        </div>
+        <AMenu ref={menuRef} {...menuComponentProps}>
+          {prependContent}
+          <div className="a-multiselect__menu-items__wrapper">
+            {!filteredItems.length && !!noDataContent && (
+              <AListItem>{noDataContent}</AListItem>
+            )}
+            {filteredItems.map((item, index) => {
```
androche, line 390: It might be nice to break all these logic blocks out to clean up the component a bit as its pretty hefty.

framework/components/AMultiSelect/AMultiSelect.scss
```diff
@@ -0,0 +1,141 @@
...
+          stroke: map-deep-get($theme, "color", "warning--state-hover");
+        }
+      }
+    }
+  }
+}
+
+.a-multiselect {
+  max-height: 116px;
+
```
androche, line 37: Before fixing this, can you take a look at https://github.com/cisco-sbg-ui/magna-react/pull/439 🙏 😃 

width

androche (changes requested):
Looks really good - notes below

framework/components/AMultiSelect/AMultiSelect.scss
```diff
@@ -0,0 +1,144 @@
...
+      .a-button:hover {
+        svg {
+          stroke: map-deep-get($theme, "color", "warning--state-hover");
+        }
+      }
+    }
+  }
+}
+
+.a-multiselect {
```
androche, line 35: `a-field-base` is set to a width of 100% so if there was a scenario where it was wrapped in a container with a flex-grow or a greater width it stretches with it.
Screenshot 2023-06-15 at 5 28 17 PM


Looks like if you do a `width:unset` here it will help avoid this stretch and doesn't change any of the new improvements you made to the width and wrapping
Screenshot 2023-06-15 at 5 28 34 PM

androche (approved):
Looks good - that was a tough one to wrangle!  One more comment below, but approving ahead of time!

PR #421, opened 2023-06-07
feat: sticky header on scroll

### LINKED ISSUE #415: Have sticky header and footer for table based pages
On screens that have a table throughout the page, make the table header and pagination footer sticky so that the table is only scrollable and is a fixed height on the screen

Examples that have sticky header + footer:
https://visibility.int.iroh.site/intelligence/indicators/public
https://visibility.int.iroh.site/snapshots

androche:
Closes #415 Adds a basic sticky header option.

@rwharrisjr  For the fixed pagination request, I feel like that is up to the table height and making sure there is scroll enabled on the table rather than the page.  What are your thoughts on that?

sticky

### DIFF
--- a/framework/components/ADataTable/ADataTable.js
+++ b/framework/components/ADataTable/ADataTable.js
@@ -105,6 +105,7 @@ const ADataTable = forwardRef(
       sort,
       selectedItems,
       truncateHeaders,
+      stickyHeader = false,
       ...rest
     },
     ref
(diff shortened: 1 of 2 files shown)


rwharrisjr:
Agreed about the footer / table scroll. I've suggested to UX that all pages with large tables be fixed with the table scroll, as it would solve a number of issues with usability.

rwharrisjr (approved):
nice!

PR #426, opened 2023-06-08
fix: display clear icon in text input based on the actual input value

### LINKED ISSUE #422: `ATextInput` ignores the clearable property
**Describe the bug**
The clearable property is now ignores. Regardless of its value, all inputs now show the clear icon once you type anything to them.

**To Reproduce**
https://github.com/cisco-sbg-ui/magna-react/blob/2cb9be6d06c55c7380c9cda2d627969618b20e3e/framework/components/ATextInput/ATextInput.js#L344

**Expected behavior**
Text input never shows the clear icon unless the `clearable` prop is set (to `true`).

**Additional context**
Regression from https://github.com/cisco-sbg-ui/magna-react/pull/377.

### LINKED ISSUE #423: `ATextInput` does not show the clear icon when re-mounted
**Describe the bug**
When `ATextInput` is re-mounted, it calls the `reset` function which de-syncs the `hasValue` state regardless of the actual value.

**To Reproduce**
1. Have a page with a (with `clearable`) `ATextInput`.
2. Type something into the input.
3. Navigate to a different page (inside the React app).
4. Navigate back.
5. The input value is preserved, but the clear icon is not shown.

**Expected behavior**
The clear icon is always visible if there is any value in the input.

**Additional context**
Regression from https://github.com/cisco-sbg-ui/magna-react/pull/377.

### LINKED ISSUE #425: `ATextInput`'s clear icon does not disapper when all input is deleted
**Describe the bug**
When all input text is manually deleted from the input, the clear icon is still visible.

**To Reproduce**
1. Have a page with a (with `clearable`) `ATextInput`.
2. Type something into the input.
3. The clear icon shows up.
4. Delete the typed input by pressing backspace.
5. The clear icon is still there.

**Expected behavior**
The clear icon disappears when the input it empty, regardless how that has been achieved.

**Additional context**
Regression from https://github.com/cisco-sbg-ui/magna-react/pull/377.

MikkCZ:
The clear icon should be displayed if and only if the value is a non-empty string, regardless of how that happened. Instead of manual syncing a separate state to determine whether the value fulfills this condition, this patch uses the value itself to determine if the clear icon should be displayed or not. It also takes the `clearable` property into account again, so the clear icon can only be shown when the text input is set to be `clearable`.

fix #422, fix #423, fix #425

### DIFF
--- a/framework/components/ATextInput/ATextInput.cy.js
+++ b/framework/components/ATextInput/ATextInput.cy.js
@@ -3,8 +3,7 @@ import ATextInput from "./ATextInput";
 const commonProps = {
   labelId: "test",
   labelFor: "test",
-  label: "test label",
-  clearable: true
+  label: "test label"
 };
 
 describe("<ATextInput />", () => {
@@ -20,7 +19,7 @@ describe("<ATextInput />", () => {
   });
 
   it("should clear input text if specified", () => {
-    cy.mount(<ATextInput {...commonProps} />);
+    cy.mount(<ATextInput {...commonProps} clearable />);
 
     // Type text
     cy.get(".a-text-input__input").type("test input");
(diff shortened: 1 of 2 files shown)


rwharrisjr (approved):
good catch on this

NeelSaini:
Verified in rel v1.123
Do not have the permission to add "QA Verified " Label

PR #428, opened 2023-06-09
fix: absolute position clear btn

### LINKED ISSUE #424: `ATextInput`'s clear icon causes the input to grow in width under certain conditions
**Describe the bug**
When the clear icon renders after typing into an input, the whole input grows in width unless explicit width is set to a common wrapping element.

**Screenshots**

https://github.com/cisco-sbg-ui/magna-react/assets/3941344/380b9502-892a-4459-8411-5cd77cca249c

**Additional context**
Regression from https://github.com/cisco-sbg-ui/magna-react/pull/377. Before the clear icon was always either present or missing, but it never was not being added/removed on-demand.

androche:
**Closes #424 - Remove padding add absolute**

absolute

### DIFF
--- a/framework/components/AInputBase/AInputBase.cy.js
+++ b/framework/components/AInputBase/AInputBase.cy.js
@@ -21,7 +21,7 @@ describe("<AInputBase />", () => {
 
     it("should invoke the clear callback on click", () => {
       cy.get(".a-input-base__clear")
-        .click()
+        .click({force: true})
         .then(() => {
           expect(mockFn.callCount).to.eq(1);
         });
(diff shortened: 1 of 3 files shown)


androche:
> Please check how it looks when there is very long text in the input. The text should not "unverflow" the close icon. When the clear button is visible, the input or a common wrapping element might need additional padding to account for the icon being displayed there.

Good call, I did miss that.

Screenshot 2023-06-09 at 12 16 37 PM

rwharrisjr (changes requested):
if you add an `appendIcon`, the absolute position overlaps 


Screenshot 2023-06-13 at 3 46 49 PM

androche:
I'm going to close this one as I think there are too many situations to account for with this approach.  It might need some jsx changes to adjust how append gets added

PR #430, opened 2023-06-12
Add incident manager setup script to gitignore

rwharrisjr:
Local test file for dev environment. Don't want to include it in the repo, so for convenience ignoring it from git

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@
 .turbo/
 
 .DS_Store
+
+test-incident-manager.sh


androche (commented):
Oh I have it named something else in my gitignore and found a nice trick to hide it - `git update-index --assume-unchanged .gitignore`.    I can change the filename if you still want to commit this though

rwharrisjr:
If we make a note about assume-unchanged I'm fine with that. Just wary that we'd want to update it in the future and it would cause a headache if we forgot about it. Could also just keep the script out of the root of the repo locally.

PR #439, opened 2023-06-14
fix: stabilize width on input fields

androche:
anotherone

#424 Trying to find the simplest solution here.  So there is a flex on the sibling `a-field-base` which I wanted to avoid because I see that being used elsewhere like checkboxes. So tried just targeting the input base class instead - Looks like it works well in all the various input scenarios that I tested with.  

avoid-flex

### DIFF
--- a/framework/components/AInputBase/AInputBase.scss
+++ b/framework/components/AInputBase/AInputBase.scss
@@ -111,6 +111,7 @@ $input-transition: border-color $transition-duration--extra-fast
 }
 
 .a-input-base {
+  display: block;
   &__surface {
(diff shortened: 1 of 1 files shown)


androche:
I do see an area in IM where it moves content into a column - however the flex should be handled by a container on the input elements, not by the inputs themselves.

As seen here, flex should be on the container here which just has margin applied.

Screenshot 2023-06-14 at 12 41 06 PM

As already handled for this table:

Screenshot 2023-06-14 at 12 44 36 PM

rwharrisjr (approved):

PR #447, opened 2023-06-22
Bump AMenuBase z-index to account for modals

rwharrisjr:
Form elements with menus (`ASelect`) inside `AModal` are opening the menu behind the modal

https://magna-react.vercel.app/components/modal

This bumps the `AMenuBase` z-index to 750, which is 50 higher than the new 700 for `APageOverlay`

@guptar8jan fyi

### DIFF
--- a/framework/changelog.mdx
+++ b/framework/changelog.mdx
@@ -14,15 +14,15 @@ The purpose of this update was to provide developers with sufficient space for t
 List of components that have a z-index set with a specific value:
 
 | Component    | Subcomponent | `z-index` |
-|:-------------|:-------------|----------:|
+| :----------- | :----------- | --------: |
 | ADatePicker  |              |       100 |
 | ADrawer      |              |       100 |
 | ASlider      |              |       100 |
 | ATimeline    |              |       100 |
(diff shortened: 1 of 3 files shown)


brennarvo (changes requested):
I was wondering why a unit test didn't catch this since there is an assertion for `AMenu` visibility within the `AModal` tests:

https://github.com/cisco-sbg-ui/magna-react/blob/5d256e6842b3c623053ce3c9e72693f6972c4522/framework/components/AModal/AModal.cy.js#L284

However, in Cypress terms, it turns out that visibility [^0] !== covering [^1]

If I add the assertion `cy.getByDataTestId("menu-item-1").click();` to that unit test, *then* it fails (which would've been a good thing when updating z-indexes).

Would you mind adding that into the unit tests as well?

[^0]: https://docs.cypress.io/guides/core-concepts/interacting-with-elements#An-element-is-considered-hidden-if

[^1]: https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Covering

guptar8jan (approved):
LGTM.

brennarvo (approved):

PR #457, opened 2023-07-06
Set line-height on ATag

rwharrisjr:
fixes scenarios when parent container sets a line-height that's bigger than the tag size


androche (approved):
Looks good!

PR #460, opened 2023-07-07
Add focus-visible to AButton; Add focus to AButtonGroup buttons

### LINKED ISSUE #459: AButtonGroup component doesn't have focus state
https://magna-react.vercel.app/components/button-group

just like regular buttons - these should probably have focus/active css styling applied

rwharrisjr:
Resolves #459 

made a guess at the focus-visible colors for dark theme, since that isn't reflected in the latest figma. The box shadow depth matches the other component set

Focused single button

Screenshot 2023-07-07 at 11 57 49 AM

"One" is focused:

Screenshot 2023-07-07 at 11 57 17 AM

the box-shadow for :focus-visible does not work well on button groups, so copied in the secondary button focus style

### DIFF
--- a/framework/components/AButton/AButton.scss
+++ b/framework/components/AButton/AButton.scss
@@ -16,10 +16,12 @@ $btn-transition: color $transition-duration--extra-fast
     $transition-timing-function--standard !default;
 
 @include theme(a-button) using ($theme) {
+  /*
   &:focus-visible {
     box-shadow: 0 0 0 2px map-deep-get($theme, "button", "focus-visible", "bg"),
       0 0 0 4px map-deep-get($theme, "button", "focus-visible", "control");
   }
+  */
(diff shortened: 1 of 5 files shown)


androche:
After looking through both dark and light modes I see how this is tricky.  lol

For dark theme its called switcher at the end and they arent grouped together like the light mode switcher.
Screenshot 2023-07-10 at 8 58 10 AM

I might suggest that we double check with UX/Design on the focus. We definitely need one, but its usually a global focus style, so it will show up on everywhere and anything in the UI that is tabbed into.
Screenshot 2023-07-10 at 8 53 29 AM

androche (approved):
Might want to add a comment in there on what we are waiting for/why its marked out

androche (approved):

PR #466, opened 2023-07-13
feat: Add ACopyButton component

### LINKED ISSUE #464: Copy To Clipboard Component
Magnetic Documentation provides designs for a `copy-to-clipboard` component that is a button with a hover tooltip & opens a popover confirmation on click/successful copy. 

A generic component all XDR MFE's can reuse for copying text to the clipboard.

We could create our own using magna-react components (`ATooltip`, `APopover`, and `AButton`) but it would not allow other teams outside of DI to use it for consistency between apps. 

**Additional context**
Note that the magnetic documentation mentioned below uses a `tertiary` button with grey stroke for the icon but in magna-react `tertiary` buttons get a blue stroke.

Magnetic References
- [Magnetic Documentation - Copy To Clipboard](https://zeroheight.com/0a43ab5cd/p/73fcef-copy-to-clipboard)
- [Figma Magnetic Components - Copy To Clipboard](https://www.figma.com/file/oVZWatImEIbl1c8sjdGxi0/%F0%9F%A7%B2--Magnetic-Design-Library?type=design&node-id=93515-113494&mode=design&t=3IJjmlo6nIzZbr06-4)

rwharrisjr:
Resolves #464 

Screenshot 2023-07-13 at 1 18 52 PM
Screenshot 2023-07-13 at 1 18 56 PM
Screenshot 2023-07-13 at 1 19 03 PM

### DIFF
--- a/framework/components/ACopyButton/ACopyButton.scss
+++ b/framework/components/ACopyButton/ACopyButton.scss
@@ -1,11 +1,5 @@
 @import "../../styles";
 
-.a-copy-button {
-  .a-icon {
-    //  margin-right: 5px;
-  }
-}
-
 .a-copy-button-tooltip.a-copy-button-tooltip--copied {
   padding: 0 !important;
   max-width: unset;
--- a/framework/components/ACopyButton/ACopyButton.scss
+++ b/framework/components/ACopyButton/ACopyButton.scss
@@ -1,0 +1,19 @@
+@import "../../styles";
+
+.a-copy-button {
+  .a-icon {
+    //  margin-right: 5px;
+  }
+}
+
+.a-copy-button-tooltip.a-copy-button-tooltip--copied {
+  padding: 0 !important;
+  max-width: unset;
+  .a-context {
+    margin: 0;
+    border-radius: $border-radius--lg;
+    border-left: 0;
+    white-space: nowrap;
+    align-items: center;
+  }
+}
--- a/framework/components/ACopyButton/ACopyButton.js
+++ b/framework/components/ACopyButton/ACopyButton.js
@@ -56,7 +56,7 @@ const ACopyButton = forwardRef(
         <AButton
           ref={ref}
           className="a-copy-button"
-          iconOnly={!children}
+          icon={!children}
           onClick={() => {
             copyToClipboard(value, containerId);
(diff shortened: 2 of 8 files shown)


framework/components/ACopyButton/ACopyButton.scss
```diff
@@ -0,0 +1,19 @@
+@import "../../styles";
+
+.a-copy-button {
+  .a-icon {
+    //  margin-right: 5px;
```
androche, line 5: Can remove

framework/components/ACopyButton/ACopyButton.scss
```diff
@@ -0,0 +1,19 @@
+@import "../../styles";
+
+.a-copy-button {
+  .a-icon {
```
androche, line 4: From mock up looks like small button should have icon width of `14px`

Screenshot 2023-07-21 at 2 38 02 PM

androche (changes requested):
One note, but looks good and works well, enter works as well.

framework/components/ACopyButton/ACopyButton.scss
rwharrisjr: Seems to use 14px when passing `small` to the component

androche: ^that was small haha

androche: Screenshot 2023-07-21 at 4 45 59 PM

rwharrisjr: Ah, on the button without text it changes the size, but not with text. I think that's an issue to sync on in AButton, outside of this PR

androche: hmm i'm not seeing that change even on the single icon only.  I think ACopy might need to accept the small prop directly rather than {...rest}.

androche: Oh its suddenly doing it now after a few refreshes.  Strange.  Ok yea let's tackle that another time!

androche: One conflict and then good to approve

androche (approved):

PR #493, opened 2023-08-17
feat: Add composable drawer components

### LINKED ISSUE #304: Add layout components to ADrawer
For easier DOM management it would be nice to have 

- ADrawerHeader
- ADrawerBody
- ADrawerFooter

`ADrawerHeader` could handle title text from a prop, as well as a close button. It could structure DOM so that text would never overlap an absolutely positioned close button.

These would be composable components similar to `AList` and `AListItem`

rwharrisjr:
Resolves #304

Adds composable components for `ADrawer`

- ADrawerHeader
- ADrawerTitle
- ADrawerSubtitle
- ADrawerBody
- ADrawerFooter

Issues resolved:

- auto content scrolling
- title text won't overlap close X button
- footer defaults to box shadow as per design, but has a `divider` option to swap back to the current style

@sabrinamokerji fyi

### DIFF
--- a/framework/components/ADrawer/ADrawer.js
+++ b/framework/components/ADrawer/ADrawer.js
@@ -221,15 +221,21 @@ ADrawer.propTypes = {
    */
   slimWidth: PropTypes.string,
   /**
+   * ***DEPRECATED*** (use same prop on ADrawerTitle)
+   *
    * Pass onClose handler for Drawer to handle onClose icon and action.
    * Default: x Icon
    */
   closeBtnOnClick: PropTypes.func,
   /**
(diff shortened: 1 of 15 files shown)


brennarvo (changes requested):
Only one comment about `ADrawerContent` would need to be addressed for our app to use these components instead, but everything else is just minor and very nit picky

brennarvo (commented):
I'm not sure how common this would be..but if somebody rendered a `ADrawer` with nothing but an `ADrawerFooter`, the footer would still be at the top:

Screenshot 2023-08-25 at 8 39 12 AM

It looks like setting the `margin-top` to `auto` would resolve that edge case.

brennarvo (approved):

PR #494, opened 2023-08-18
ADataTable arrow up/down keys support

bredy:

New feature witch supports: [#1157](https://github.com/advthreat/incident-manager/issues/1157)

There is no such feature.

Adds support for moving in ADataTable using arrow up/down keys.

No. It should be backward compatible. Only using new property enables it.

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -70,12 +70,12 @@ Cypress.Commands.add("getByAriaLabel", (label) => {
 });
 
 const keydownUtils = [
-  {keyCode: keyCodes.esc, name: "escapeKeydown"},
-  {keyCode: keyCodes.enter, name: "enterKeydown"},
-  {keyCode: keyCodes.space, name: "spaceKeydown"},
-  {keyCode: keyCodes.down, name: "downArrowKeydown"},
-  {keyCode: keyCodes.up, name: "upArrowKeydown"},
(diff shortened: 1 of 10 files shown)


rwharrisjr:
UX wise, I think it's more intuitive if when the table loses focus, the starting point of the next focus/keyboard resets to the first traversable element. 

For example, the "Form Example" on https://magna-react-nfdg8iims.securex-preview.app/components/modal. Clicking outside a form input resets the tab iterator to the first focusable element of the modal.

---------

In the case there are two keyboard traversable tables on the same page, the up/down event fires in both tables

Screenshot 2023-08-23 at 12 50 40 PM

bredy:
> UX wise, I think it's more intuitive if when the table loses focus, the starting point of the next focus/keyboard resets to the first traversable element.

I have currently this functionality in place where I am using data table, in incident-manager. Thought it does not belong to lib. I can move it to magna as well. 

> In the case there are two keyboard traversable tables on the same page, the up/down event fires in both tables

Yes, I know about it, was already thinking about it and decided not to deal with that in this moment and let user dynamically turn on/off keyboard support for each table he would like to currently support. Frankly, how it should behave when 2 tables should support page navigation in the same moment ?

rwharrisjr (approved):
very cool!

PR #498, opened 2023-08-22
fix: Support focus in ACheckbox (#497)

### LINKED ISSUE #497: Missing focus on "ACheckbox"
**Describe the bug**
Currently there is no style for `ACheckbox` "focus".

**To Reproduce**
Steps to reproduce the behaviour:
1. Check first checkbox in "[Usage](http://localhost:8001/components/checkbox#usage)" 
2. Click "Tab" to go to second checkbox
3. No visual shows me second checkbox is focused.
-->

**Expected behaviour**
Second checkbox having focus is visually different compared to others - it should have box-shadow according to magnetic.

bredy:

Fix, closes #497

No, it should be backward compatible.

**Other information**:
It would be useful to replace unchecked SVG with right size SVG.

### DIFF
--- a/framework/components/ACheckbox/ACheckbox.js
+++ b/framework/components/ACheckbox/ACheckbox.js
@@ -156,10 +156,14 @@ const ACheckbox = forwardRef(
     }
 
     let currentPath = Icons.unchecked;
+    // TODO: "unchecked" svg is smaller (19.6px) I need to update scale, replace it with svg of same size as "indeterminate" and "checked" (20px)
+    let currentViewBox = "0 0 14.6 14.6";
     if (indeterminate) {
       currentPath = Icons.indeterminate;
+      currentViewBox = "0 0 15 15";
     } else if (checked) {
       currentPath = Icons.checked;
+      currentViewBox = "0 0 15 15";
     }
 
     const handleKeyDown = (e) => {
@@ -171,7 +175,7 @@ const ACheckbox = forwardRef(
     };
 
     const checkboxContent = (
-      <>
+      <div className="a-checkbox__focuswrap">
         <input
           id={id}
           type="checkbox"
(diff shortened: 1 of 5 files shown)


bredy:
If anybody provides me "unchecked" SVG of correct size I can update is in PR as well.

bredy:
Screenshot 2023-08-22 at 15 38 12
Screenshot 2023-08-22 at 15 38 31
Screenshot 2023-08-22 at 15 38 49

rwharrisjr:
CDS has the focus state as changing the fill of the svg. I think we can do that with the :focus selector here instead of the wrap. (the dark theme is barely distinguishable for focus...). Top checkbox is focused in both images

Screenshot 2023-08-22 at 9 26 41 AM

Screenshot 2023-08-22 at 9 25 02 AM

bredy:
See "CTR UX/UI" channel please - there is proposed focus style from Alex:
Screenshot 2023-08-21 at 1 36 07 PM
Screenshot 2023-08-21 at 1 37 12 PM

I think we need to style using "border-shadow" and to do it we need wrapped element - especially because we can utilise "focus-within". I think there is no other way how to keep native focus behaviour of input "hidden" behind icon.

rwharrisjr:
I see that's what the magnetic ux is suggesting, even though it's not defined in documentation. It conflicts with the CDS kit, which is what we are typically supposed to align to. 

We can go with this. I wouldn't be surprised if the design (or CDS) changes on us though.

rwharrisjr:
Also note I did try this sort of focus on button to match magnetic-react, but the UX wasn't great on button groups. Here it does work a lot better since checkboxes are never directly adjacent.

rwharrisjr:
@bredy focus state shows the check on checkboxes without a label in unchecked state, and no check in the checked state.

Screenshot 2023-08-22 at 10 35 21 AM

bredy:
@rwharrisjr I have updated the example - in component there is `onKeyDown` handler attached on wrapper `div` which triggers `onClick` callback ... not very intuitive ... every `onClick` handler should check what is the source of the event - if he wants just clicks. It would be maybe useful to split these callbacks and have isolated `onClick` and `onKeyDown` but it would break backward compatibility.

rwharrisjr:
Minor size issue, but I don't think anyone should be using the `small` prop for checkboxes at this point.

Screenshot 2023-08-22 at 12 21 52 PM

rwharrisjr (approved):
Lgtm. I'll address the keyboard input on the no label checkbox in a subsequent PR

PR #512, opened 2023-09-08
CSS file through webpack

androche:
**Adding another option to consider using webpack.**  

This builds a single, compiled css file to `lib/styles/magna-react.css` (but can be anywhere) which we can utilize to **1)** add to DOM as seen in the rollup PR.   Note - Webpack offers an insert link option through the `mini-css-extract-plugin` - https://webpack.js.org/plugins/mini-css-extract-plugin/, but that appears to only be use for applications rather than a component library.  Worth double checking that though as that is what we are trying to do. 

Other ideas for use of this might be a more common use of **2)** surfacing `/dist` and allowing applications to bring in their on import '@dist/magna-react.css'` and/or **3)** stick in shell and have the individual applications to conditionally use like they already do for local dev.  Seems like magna controlling DOM manipulation it would be less complicated though.  More likely apps would probably get "CSS surprises" once going to shell if they are behind on versions.

**Update for troubleshooting** - Was unable to get es modules to work in IM for `/dist/bundle.js`  Various errors in production builds were potentially linked to next.js, webpack 5, and/or babel issues.  

Screenshot 2023-09-14 at 3 15 36 PM


androche:
Closing for now as webpack esmodules support was just merged in May 2023 as "experimental"  https://github.com/web-infra-dev/rspack/issues/2293 , however doesn't seem to be working for us.

androche:
Closing for now as webpack esmodules support was just merged in May 2023 as "experimental" `webpack web-infra-dev / rspack / issue / 22932293`, however doesn't seem to be working for us.

PR #522, opened 2023-09-14
feat: add bold text to matched letters

### LINKED ISSUE #491: Lookahead search component should bold search

Per https://magnetic.cisco.com/0a43ab5cd/p/42131c-select/t/9911aa we should highlight matching search text in the lookahead search component. 

We used the small [fuzzysort](https://www.npmjs.com/package/fuzzysort) package which supports highlighting to in https://github.com/advthreat/incident-manager/pull/1299 for one of our search components and we think the same could be done for the magna component so the solution can be used elsewhere.

We could also tear out the parts of the fuzzysort package and rewrite what we need to make our own implementation and not add another dependency

**Additional context**

androche:
**Resolves #491 - Add bold text to search**

Was a little tricky with mixing regex with jsx, but this seems to work for most purposes. Would need to be improved for the object items however.

regex

### DIFF
--- a/framework/components/AAutocomplete/AAutocomplete.js
+++ b/framework/components/AAutocomplete/AAutocomplete.js
@@ -121,6 +121,21 @@ const AAutocomplete = forwardRef(
       setError("");
     };
 
+    const handleBoldText = (item) => {
+      if (!value) {
+        return item;
+      }
+      const regex = new RegExp(`(${value})`, "gi");
+      const itemArray = item.split(regex);
+      const boldText = itemArray.map((str, i) => {
+        if (regex.test(str)) {
+          return <b key={`bold-item-${i}`}>{str}</b>;
+        }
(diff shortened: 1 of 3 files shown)


framework/components/AAutocomplete/AAutocomplete.js
```diff
@@ -121,6 +121,21 @@ const AAutocomplete = forwardRef(
       setError("");
     };
 
+    const handleBoldText = (item) => {
+      if (!value) {
+        return item;
+      }
+      const regex = new RegExp(`(${value})`, "gi");
```
rwharrisjr, line 128: nice

rwharrisjr (changes requested):
Seems to work well, just needs to be added to AMultiSelect too

rwharrisjr (approved):
nice!

PR #526, opened 2023-09-23
525 - swap to use padding, adjust a bit of spacing

rwharrisjr:
quick pass at swapping padding for margin

### DIFF
--- a/framework/components/APageLayout/APageHeader.scss
+++ b/framework/components/APageLayout/APageHeader.scss
@@ -1,5 +1,5 @@
 .a-page-header {
-  margin: 0 0 24px 0;
+  padding: 0 0 24px 0;
   line-height: 34px;
 
   .a-breadcrumb {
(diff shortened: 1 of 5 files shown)


framework/components/APageLayout/APageHeader.scss
```diff
@@ -1,18 +1,18 @@
...
+  padding: 0 0 24px 0;
   line-height: 34px;
 
   .a-breadcrumb {
     padding: 0 !important;
   }
 
   .a-page-title {
-    margin: 24px 0;
+    padding: 0 0;
```
androche, line 10: I would have magna manage the 24px on the bottom here.
Screenshot 2023-09-25 at 9 15 19 AM

androche (approved):
I was able to test locally in IM up to a certain point, but then it started saying the imports cant be found.  Other than that so far so good.

PR #533, opened 2023-09-27
remove color css on checkbox

rwharrisjr:
Conflicts with old styles, and isn't necessary anymore I think

--------

Should darken on hover for default, warning, and danger checkboxes - default darkens on hover for magnetic. Should lighten on custom closers - no magnetic equivalent and this keeps it non breaking


androche (approved):
🤞

androche (approved):

PR #534, opened 2023-09-27
Remove first and last of type margin override

rwharrisjr:
Causes issues in tabs that are wrappered, such as ones with `<a>`.


androche (approved):
Pulled and tested
Screenshot 2023-09-27 at 4 47 46 PM

PR #542, opened 2023-10-04
Default toast placement to upper right

rwharrisjr:
Magnetic says toasts should go in the upper right, "24px inset below the header and right edge of the browser window."

Before

Screenshot 2023-10-04 at 3 31 23 PM

After

Screenshot 2023-10-04 at 3 31 01 PM

Note that the top offset accounts for the 56px of the XDR header, so it looks a bit off on the magna site


androche:
If you happened to have a before after that would be helpful for reference here as well :)

androche (approved):

PR #561, opened 2023-11-03
feat: initial commit and proposed cds token integration 

### LINKED ISSUE #549: Potential conversion to CDS color tokens
**Overview**
We've been asked by UX if Magna might eventually utilize CDS color tokens.  Reasons for prior hesitancies were that dark mode didn't exist, churn as magnetic was developing and iterating, as well as requirements for atomic to magnetic transition and support.  Now that CDS has released dark mode, this option can be revisited.  Design has also now fully adopted CDS color tokens as well so there's less concern for churn moving forward.

**Impact on other teams**
-Monorepo is adopting palette so we need to follow up as soon as we make a decision. https://github.com/advthreat/securex-ui-monorepo/issues/1425
-All of ribbon would need to be migrated off palette.

**Step 1: Research**
- Feasibility discussion set for week of 10/23/23
- Attendees @rwharrisjr and @androche

androche:
**Resolves #549 - CDS magnetic-tokens and theming integration**

**Feature**
CDS design-tokens provides several formats for use, in this case we are utilizing scss variables to integrate into our current theming structure.  This will also allow other groups to utilize the scss dark and light palettes as well.  The other benefit to this approach is there is no need to namespace since we are not duplicating shell app's css variables.

**Utilization** 
Seen in the `Card` example below: 

`"container-background": map-get($mds-neutral, "neutral-16")`
to
 `"container-background": $base-bg-weak-default`

What is also nice about this is that it can be consolidated to 1 theme file. 

**Maintenance**
~~Clone CDS and generate a script that copies over and compresses the files as seen in this PR and shell app.  We can then easily run a command such as `npm run update_CDS` when there is a new release.~~ After some discussion with the CDS group, they would prefer us to utilize artifactory and are ok with this token package being public (MIT license).  This will help with maintenance, although we will have to configure our CI for this approach.

**Potential pitfalls**
Inevitably, there will be updates such as the most recent breaking change with naming conventions from CDS.  In these cases, we would would need to make these same adjustments in the theming files.  This could potentially break all other apps using their palettes as well, but that is in line with any group that consumes CDS.  We will just have to pay attention to any announcements. 

**Outstanding questions**
1. Do we see this approach causing any issues in our current build process.
2. Are there any other pitfalls not considered here that would require abandoning or adjusting this plan?


framework/components/AStepper/AStep.js
```diff
@@ -12,7 +12,7 @@ const AStep = forwardRef(
       active,
       visited,
       disabled,
-      showIconOnVisited,
+      showIconOnVisited = true,
```
rwharrisjr, line 15: Note this

rwharrisjr (approved):
WOOOHOOO!!!!!

PR #581, opened 2023-12-18
feature: ADataTable takes header properties and pass them to ADataTableHeader ("td") element

bredy:

- supports: [#2047](https://github.com/advthreat/incident-manager/issues/2047)
- be able to add attributes which will be rendered on header elements
- split very complex ADataTable code to logical components which should help when adding new features

- there is no possibility to pass any attribute to ADataTableHeader level.

- we can add properties which are used while rendering ADataTableHeader elements.

- no, this feature was created to maintain backward compatibility

### DIFF
--- a/framework/components/ADataTable/ADataTable.cy.js
+++ b/framework/components/ADataTable/ADataTable.cy.js
@@ -32,6 +32,31 @@ const ITEMS = [
 ];
 
 describe("ADataTable", async () => {
+  it("renders data-testid in the table header elements", () => {
+    const headersWithProperties = [
+      {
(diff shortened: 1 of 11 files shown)


androche (commented):
Thanks for taking all of this on @bredy! 🎉  This is all very thorough - I've been clicking around for a while now and am not seeing issues.   @rwharrisjr while testing I noticed we do have some ongoing issues with keyboard not selecting checkboxes or expanding rows so we could open tickets to look into those another time.

rwharrisjr (approved):
everything seems to work as before. Great refactor!

PR #587, opened 2024-01-05
feat: divider additions

### LINKED ISSUE #586: Add 1px option to ADivider
Most documentation have 1px dividers rather than 2px which is helpful for some teams using ADivider as a border (brought up by Raj).  2px still exists in some mockups so we will keep that as default. CDS also has 2 colors, one light (current) and one dark so we should add those as well.

- Keep 2px as default
- Covert `light` variant to 1px
- Add prop for darker color

androche:
Closes #586 - Adds options to divider

- No breaking changes to default (2px)
- `Light` variant now refers to height which will be 1px
- Adds option for `strong` variant which is a darker color

Screenshot 2024-01-05 at 12 29 25 PM

### DIFF
--- a/framework/components/ADivider/ADivider.scss
+++ b/framework/components/ADivider/ADivider.scss
@@ -3,8 +3,8 @@
 @include theme(a-divider) using ($theme) {
   border-color: map-deep-get($theme, "divider", "color-default");
 
-  &--color-light {
-    border-color: map-deep-get($theme, "divider", "color-light");
+  &--strong {
+    border-color: map-deep-get($theme, "divider", "color-strong");
   }
 }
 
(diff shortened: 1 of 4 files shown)


framework/components/ADivider/ADivider.scss
```diff
@@ -13,4 +13,8 @@
   border-width: $border-width 0 0 0;
   border-style: solid;
   margin: 20px 0;
+
+  &--light {
+    border-width: 1px 0 0 0;
```
rwharrisjr, line 18: Could this be `thin` instead of `1px`? At one point browser scaling would remove 1px borders but `thin` would still show up as expected. Might not be and issue anymore

androche: Oh I haven't heard of that one - seems ok, but will `thin` always stay 1px?  Or is it relative to something

rwharrisjr: It used to always stay 1px and display nicely during scaling. Probably worth trying out again to see what browsers currently do with it

rwharrisjr (approved):

PR #595, opened 2024-01-16
Add generic prepend and append content to ATextInput

### LINKED ISSUE #594: Add ability to append text to end of input box like icons
Wish to be able to append small text to the end of an input box, similar to icons. 

UX has stated that when possible they would like to see text instead of icons for certain form elements. In this case we wish to add the text "Show" to the end of a password input so users can click to see / hide their password as they are typing it. An example is found in Cloud Sign-On.

IMG_7090

Just using an icon.

rwharrisjr:
Resolves #594 

Also fixes padding issues inside the input (always added `append` to AInputBase, causing scss from ATextInput.scss to have padding 0)

### DIFF
--- a/framework/components/AInputBase/AInputBase.js
+++ b/framework/components/AInputBase/AInputBase.js
@@ -34,12 +34,19 @@ const AInputBase = forwardRef(
     },
     ref
   ) => {
+    const showPrepend =
+        (Array.isArray(prepend) && prepend.length) ||
+        (!Array.isArray(prepend) && prepend),
+      showAppend =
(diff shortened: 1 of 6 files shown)


androche (commented):
Works well!  Although I think 99% of all cases would be a button which would take care of the padding, everything else ends up right at the edge though.  Should we handle that or is it something users should take care of on their end.
Screenshot 2024-01-17 at 9 40 42 AM

Also we may want to add a bit more of a description behind why someone would want to use appendIcon vs append generic with an icon since that might be confusing at first glance.

androche (approved):

PR #613, opened 2024-02-07
Add AKeyValue components

rwharrisjr:
https://magnetic.cisco.com/0a43ab5cd/p/31cdc6-key-value-pairs/b/417681

### DIFF
--- a/framework/components/AKeyValue/AKeyValue.js
+++ b/framework/components/AKeyValue/AKeyValue.js
@@ -1,0 +1,66 @@
+import React from "react";
+import PropTypes from "prop-types";
+import AIcon from "../AIcon";
+import ATriggerTooltip from "../ATriggerTooltip";
+import "./AKeyValue.scss";
+
+const baseClass = "a-key-value";
+
+const AKeyValue = ({
+  label,
+  children,
+  tooltip,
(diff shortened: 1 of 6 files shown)


framework/components/AKeyValue/AKeyValue.js
```diff
@@ -0,0 +1,66 @@
...
+import ATriggerTooltip from "../ATriggerTooltip";
+import "./AKeyValue.scss";
+
+const baseClass = "a-key-value";
+
+const AKeyValue = ({
+  label,
+  children,
+  tooltip,
+  icon = "information",
```
androche, line 13: Size on the icon is a little large
Screenshot 2024-02-07 at 6 03 51 PM

androche (approved):
👍  The only thing i kept looking back at is the position of the tooltips. They are almost so slightly lower than I would expect and also seems to vary between font sizes maybe.  Very minor, but maybe if there's a way to bump them up 1px i think that might help?
Screenshot 2024-02-08 at 9 25 56 AM

rwharrisjr:
I adjusted the line heights a bit which kinda helps, but aligning icons against text is always wonky

androche (approved):

PR #614, opened 2024-02-08
feat: field base can have multiple hints

bredy:
**Please check if the PR fullfills these requirements**

UX needs from us to display field hint and optionally display error message - but keeps hint visible (see screenshot). This was not possible to reach with "hint" and "rules" because "rules" needs validation and validation needs user interaction. But we need to control both properties from outside because we are using external validation framework.

I am adding new property "hints" allowing me to put one or more hints to "AFieldBase" and related components.  Every hint can be configurable, we can set alarm level, we can hide it when validation error occur, we can infer component validation status. I kept there original "hint" for backward compatibility and marked it as deprecated. I also marked as deprecated on "AFieldBase" related properties "hintUsesValidationState" and "hideHintOnError".

Supports: https://github.com/advthreat/incident-manager/issues/2228

Current behaviour supports only mix of single hint combined with validation error.

This change allows to put more hints to "AFieldBase" and related components.  

No, change should be backward compatible.

### DIFF
--- a/framework/components/AFieldBase/AFieldBase.js
+++ b/framework/components/AFieldBase/AFieldBase.js
@@ -99,7 +99,7 @@ const AFieldBase = forwardRef(
           </label>
         )}
         {children}
-        {hints &&
+        {Array.isArray(hints) &&
           hints.map((hintObject, index) => {
             // "hints" block should be before "error" hint
             if (hintObject.hideHintOnError && error) return null;
--- a/framework/components/AFieldBase/AFieldBase.js
+++ b/framework/components/AFieldBase/AFieldBase.js
@@ -14,6 +14,7 @@ const AFieldBase = forwardRef(
       className: propsClassName,
       error,
       hint,
+      hints,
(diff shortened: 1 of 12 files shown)


framework/components/AFieldBase/AFieldBase.js
```diff
@@ -97,6 +99,27 @@ const AFieldBase = forwardRef(
           </label>
         )}
         {children}
+        {hints &&
```
androche, line 102: May want to check length on here for safety

androche (commented):
Pulled and tested - Works really well!  Passing to @rwharrisjr for approval

rwharrisjr (approved):
lgtm

framework/components/AFieldBase/AFieldBase.js
bredy: @androche  Thank for the review.

I rather replaced it with `Array.isArray` to check if we are really iterating over array. If array is empty it just produces nothing - so it should be fine.

PR #621, opened 2024-02-14
feature: accordion title has now way to place chevron icon to the left

bredy:

Enables to configure placement of `AAccordionHeaderTitle` icon. Now we have way to place it also to the left side.
We verified with UX that they really would like to have it left.

Supports [#1539](https://github.com/advthreat/incident-manager/issues/1539)

Currently we are able to place the icon only to the right side.

No this change is backward compatible.


rwharrisjr (approved):
lgtm

PR #624, opened 2024-02-22
feat: add "gap" css helper class

brennarvo:
## Context

When rendering a flexbox container, it is often necessary to add individual `ml-*` or `mr-*` classes to each list item to introduce a space between flex items. However, when doing so, there is often a need to conditionally apply the class such that this class is **not** applied to the very first flex item when using a `margin-left`, or alternatively, the last flex item when using `margin-right`.

Such conditional logic can be avoided with the [`gap` CSS shorthand property](https://developer.mozilla.org/en-US/docs/Web/CSS/gap), which will only add the space *between* adjacent elements.

That being said, there has been some hesitation for using this within applications since we can't use the design system spacing sizes directly in our CSS with the gap property, e.g:

```scss
.container {
  display: flex;
  gap: 5px; // hardcoded; not using magna react spacing units
}
```
Though @androche @rwharrisjr I want to say this is now possible with CSS variables introduced by Magna, correct? E.g. `gap: var(--space-sm);` or something similar?

Regardless, it would be nice to have these spacing units available via a helper class for `gap`

## This PR

* Adds a helper class so that we can use the design system spacing as a helper class for the `gap` property
  * Sample usage:
    ```html
    <div className='d-flex gap-1'>
      <AButton>Foo</AButton>
      <AButton>Bar</AButton>
      <AButton>Baz</AButton>
    </div>
    ```
* Adds includes the token generation step as part of the local development instructions for running the application locally

## Test Plan

1. Navigate to [the button component playground](https://magna-react-8sr970s1z.securex-preview.app/components/button)
2. Paste the following into the playground editor:
    ```html
    <div className='d-flex gap-1'>
      <AButton>Foo</AButton>
      <AButton>Bar</AButton>
      <AButton>Baz</AButton>
    </div>
    ```
3. Change the number appended to the `gap-` classname to see the spacing differences (I think the largest size we support is `12`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ Feel free to reach out to Robert Harris (roberha2) via Webex Teams to discuss co
 ### Run Locally
 ```
 npm ci
+npm run addTokens
 npm run dev
 ```
 
--- a/README.md
+++ b/README.md
@@ -70,7 +70,6 @@ Feel free to reach out to Robert Harris (roberha2) via Webex Teams to discuss co
 ### Run Locally
 ```
 npm ci
-npm run addTokens
 npm run dev
 ```
 
--- a/framework/components/AApp/base/helpers.scss
+++ b/framework/components/AApp/base/helpers.scss
@@ -63,6 +63,18 @@ $helpers: (
     class: flex,
     values: wrap nowrap wrap-reverse
   ),
+  "gap": (
+    responsive: true,
+    property: gap,
+    class: gap,
+    values:
+      map-merge(
+        $spacers,
+        (
+          auto: auto
+        )
+      )
+  ),
   "justify-content": (
     responsive: true,
     property: justify-content,
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "husky:pre-commit": "lint-staged",
     "prepare": "husky install",
     "publish-package": "rimraf lib && npm ci && npm run build && npm publish",
-    "dev": "next dev -p 8001",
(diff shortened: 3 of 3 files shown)


framework/components/AApp/base/helpers.scss
```diff
@@ -63,6 +63,18 @@ $helpers: (
     class: flex,
     values: wrap nowrap wrap-reverse
   ),
+  "gap": (
+    responsive: true,
```
brennarvo, line 67: To be completely honest, I am not sure what the implications of `responsive` here are, but most of the other helper classes (particularly the ones for flexbox) are using it 🤷

README.md
```diff
@@ -70,6 +70,7 @@ Feel free to reach out to Robert Harris (roberha2) via Webex Teams to discuss co
 ### Run Locally
 ```
 npm ci
+npm run addTokens
```
androche, line 73: Good catch!  We could probably handle it in the package.json to avoid an extra step for ppl -  would you mind adding to package.json line 30 `"dev": "npm run addTokens && next dev -p 8001",`.

androche (commented):
Looks good!  
> I want to say this is now possible with CSS variables introduced by Magna, correct? E.g. gap: var(--space-sm); or something similar? 

Yea we are doing a bunch of spring cleaning of the old css -  I didn't even think about extending other vars out in that sense so we will definitely noted for those improvements!

README.md
brennarvo: Good idea. Done!

PR #629, opened 2024-02-26
Rewrite cds token class groups to isolate them from the CDS package

rwharrisjr:
Allows magna-react to provide the tokens on the `.theme-*` classes that it already uses, without any conflicts with the same tokens provided by any other library.

This shouldn't cause any more duplication than already exists, and allow us to remove the `data-cds-theme="..."` on `ATheme` and any custom implementations


framework/components/ATheme/ATheme.js
```diff
@@ -129,7 +129,7 @@ const ATheme = forwardRef(
     }
 
     return (
```
androche, line 131: Can delete all dataTheme references

androche (approved):
🎉 🎉 !!

androche (approved):

PR #635, opened 2024-03-04
fix-627 Add accessibility features to modals and drawers

### LINKED ISSUE #627: [Accessibility] Modals and Drawers don't keep focus placement on close
When a user clicks on an element which opens either `AModal` or `ADrawer` components, the focus should return to that initial element upon closing said modal or drawer. Instead, the user must tab all the way from the start of the page to get back to where they were.

Open a Modal or Drawer, then close it and hit tab to see that the focus is at the top of the page. 

**Expected behavior**
Closing a modal or drawer should place the focus back on the element used to open them. It's outlined here in this example under "End of the Road!" section -
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/

meggettt:
- [ ] The commit message follows semantic commit message guidelines
- [ ] Has been verified locally

Closes  #627 

The current behavior is outlined in the issue here - #627 

When closing modals and drawers, the focus should return to whatever element was used to them. Also, they should be able to be closed upon pressing the "Escape" key, provided that an "onClose" function is passed into their props. The usePopupQuickExit hook is utilized to give these components this ability out of the box. 

Although not a breaking change, an onClose function to both Modals and Drawers. 

**Other information**:

### DIFF
--- a/framework/components/AModal/AModal.js
+++ b/framework/components/AModal/AModal.js
@@ -15,6 +15,7 @@ import {
 
 import "./AModal.scss";
 import {useDelayUnmount} from "../../utils/hooks";
+import usePopupQuickExit from "../../hooks/usePopupQuickExit/usePopupQuickExit";
 
 /**
  * Reasons for not stopping propagation for the following keys:
@@ -39,11 +40,13 @@ const AModal = forwardRef(
       delayUnmount = 200,
       children,
       isOpen: propsIsOpen,
+      onClose,
       small,
       medium,
       large,
       xlarge,
-      onClickOutside,
+      onClickOutside, //***DEPRECATED***
+      closeOnOutsideClick = false,
       withCenteredContent = true,
       withFocusTrap = true,
       autoFocusElementRef,
@@ -59,6 +62,7 @@ const AModal = forwardRef(
     const hasPortalNode = appendTo || appRef.current;
     const isOpen = !!hasPortalNode && propsIsOpen;
     const _ref = useRef();
+    const childRef = useRef();
 
     const shouldRenderChildren = useDelayUnmount({
       isOpen,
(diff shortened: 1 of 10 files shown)


framework/components/AModal/AModal.js
```diff
@@ -298,9 +310,16 @@ You should provide either an \`aria-label\` or \`aria-labelledby\` prop to \`${c
...
    * If not using the `usePopupQuickExit` hook, pass in a function to handle clicking outside of the modal event. `withOverlay` prop must be set to true.
+   * @deprecated use "closeOnOutsideClick" in conjunction with a passed in "onClose" function to close modal
    */
   onClickOutside: PropTypes.func,
 
+  /**
+   * Function which closes the Modal. It is necessary for accessibility concerns, 
+   * specifically to enable exiting the Modal upon pressing the "Escape" key.
+   */
+  onClose: PropTypes.func,
```
rwharrisjr, line 321: Might consider making this required, since that'll help enforce accessibility.

framework/components/AModal/AModal.js
```diff
@@ -39,11 +40,13 @@ const AModal = forwardRef(
       delayUnmount = 200,
       children,
       isOpen: propsIsOpen,
+      onClose,
       small,
       medium,
       large,
       xlarge,
-      onClickOutside,
+      onClickOutside, //***DEPRECATED***
```
rwharrisjr, line 48: since the function isn't used anymore, it might be better to remove it entirely and put in a warning

```
if (onClickOutside) {
console.warn("onClickOutside has been removed. Use onClose and set closeOnOutsideClick to true") 
}
```

If we make `onClose` required as well, this would add a bit more guidance to help people update

framework/components/AModal/AModal.mdx
```diff
@@ -16,26 +16,26 @@ import {AModal} from "@cisco-sbg-ui/magna-react";
 
 ## Usage
 
+<AAlert className="mb-2" level="info" dismissable={false}>
+  Please provide an "onClose" function for accessibility concerns, specifically to enable exiting
```
rwharrisjr, line 20: If we make `onClose` required, can remove the "please" :)

rwharrisjr (changes requested):
A few changes.

Also appears to be failing the cypress tests

rwharrisjr (approved):
lgtm!

PR #639, opened 2024-03-06
Allow onKeyDown to activate edit mode on inline inputs

rwharrisjr:
Has an onKeyDown, but wasn't set to activate edit mode


androche (approved):
Tested in deployment

PR #642, opened 2024-03-13
Keyboard accessibility for ASwitch

rwharrisjr:
Adds focus state and keyboard toggling on "enter" key for `ASwitch`


androche (approved):
👍

rwharrisjr:
The `checked` attr doesn't get passed down to the DOM from react.. so leaving `aria-checked` for now. Added a note about aria labels.

meggettt (approved):

PR #644, opened 2024-03-14
Show tooltip on focusin; close on escape

rwharrisjr:
Accessibility

> The tooltip should appear on focus or when the element is hovered on, without additional interaction. It should disappear automatically when the focus on the owning element is lost or the mouse is moved outside the owning element and the tooltip. While the tooltip does not receive focus, the Escape should close it if it is open.

Only applies to `ATriggerTooltip`


androche (approved):

androche (changes requested):
Oops - actually I tested on tooltip and couldnt get a focus going.  Can we add this over there too?

rwharrisjr:
The events for `ATooltip` are defined outside of the component, so the focus events would need to be defined at that point as well.

androche (approved):

PR #650, opened 2024-03-20
fix-646 Add directional navigation capabilities

### LINKED ISSUE #646: TabGroup Accessibility
Tabs have specific accessibility requirements dealing with how they are accessed. 
When a Tabgroup receives focus, the focus automatically goes to the active tab. 
When a user presses tab again, the focus goes from the active tab to the next tabbable element outside of the Tabgroup. 
Pressing the left and right (or up and down) keys cycles through the tabs
The tabs are automatically activated while being cycled through. 
Additional aria labels are added 

The "More" tab is going to take a bit of reworking, especially considering the bug where two tabs can be highlighted when the more menu is active. 

https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

meggettt:
- [ ] The commit message follows semantic commit message guidelines
- [ ] Has been verified locally

Closes #646 

Tabs cannot be navigated using keyboard inputs as Accessibility concerns demand.

Using directional keys will cycle through the tabs and automatically open them.

More Tab functionality still needs to be worked out.

### DIFF
--- a/framework/components/ATabs/ATab.cy.js
+++ b/framework/components/ATabs/ATab.cy.js
@@ -78,23 +78,18 @@ describe("<ATabGroup />", () => {
       .contains("Seven");
   });
 
-  it("should maintain highlighted state if overflow menu item is selected by keystroke", () => {
-    cy.mount(<ATabTest width={"15rem"} />);
+  it("should automatically open tab when selected by keystroke", () => {
+    cy.mount(<ATabTest width={"15rem"} uncontrolled={true} />);
     cy.get(".a-tab-group__tab")
       .first()
-      .tab({shift: true})
-      .contains("More")
-      .type("{enter}")
+      .focus()
+      .type("{rightArrow}")
+      .contains("One")
       .get(".a-tab-group__tab--selected")
-      .should("exist");
-    cy.get(".a-list-item").first().next().type("{enter}");
-    cy.get("body").click(0, 0);
-    cy.get(".a-tab-group__tab--selected").should("exist");
-    cy.get(".menu-tab").click();
-    cy.get(".a-list-item")
-      .get(".a-list-item--selected")
       .should("exist")
-      .contains("Four");
+      .type("{leftArrow}")
+      .get(".tab-overflow-menu")
+      .should("exist");
   });
 
   it("should not render overflow tab if in vertical position", () => {
@@ -103,7 +98,12 @@ describe("<ATabGroup />", () => {
   });
 });
 
-const ATabTest = ({width = "30rem", vertical = false, secondary = false}) => {
+const ATabTest = ({
+  width = "30rem",
+  vertical = false,
+  secondary = false,
+  uncontrolled = false
+}) => {
   const tabs = ["One", "Two", "Three", "Four", "Five", "Six", "Seven"];
   const [activeTab, setActiveTab] = useState(tabs[2]);
   return (
@@ -112,7 +112,7 @@ const ATabTest = ({width = "30rem", vertical = false, secondary = false}) => {
         {tabs.map((x) => (
           <ATab
             key={x}
-            tabKey={x}
+            tabKey={!uncontrolled && x}
             onClick={() => setActiveTab(x)}
             selected={x === activeTab}
           >
--- a/framework/components/ATabs/ATab.cy.js
+++ b/framework/components/ATabs/ATab.cy.js
@@ -78,7 +78,7 @@ describe("<ATabGroup />", () => {
       .contains("Seven");
   });
 
-  it("should automatically open tab when selected by keystroke", () => {
+  it("should cycle through tabs by left and right keystroke", () => {
     cy.mount(<ATabTest width={"15rem"} uncontrolled={true} />);
     cy.get(".a-tab-group__tab")
       .first()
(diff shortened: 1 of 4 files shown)


androche:
@meggettt Tabs are a bit tricky due to the dynamic more overflow and toggling of classes there.  While trying to test though it doesnt appear to be engaging in tab through tabbing

meggettt:
> @meggettt Tabs are a bit tricky due to the dynamic more overflow and toggling of classes there. While trying to test though it doesnt appear to be engaging in tab through tabbing

The tabs aren't supposed to actually be able to be engaged with through tabbing - although we probably do need UX to give some direction on how to highlight the selected tab once it's focused. 

When you tab into to a tabGroup, only the active tab should receive focus. Then the user should be able to use the navigation keys to cycle through the tabs and automatically open them. For the sake of just checking the functionality, that can be accomplished in the preview even if there isn't a visual indicator that the user is focused on that active tab yet.

meggettt:
> @meggettt thanks for clarifying that! Yea something about not having tabs was a bit wonky for me trying to navigate - maybe just habit. Not sure if its ok to do both or if that complicates things. Or I wonder if we tabbed in with a focus on the whole container first and then arrow in my call it out a bit more. Not sure i'll have to pass all that over to @rwharrisjr for opinions, but with all that being said initially I was trying on the "controlled mode" and it seemed like it was not focusing at all for arrows either.
> 
> Screenshot 2024-04-02 at 11 36 16 AM

It is a bit unintuitive for those who are used to using tabbing, but it's an accessibility standard -https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ 

If users are using Controlled mode then they will have to roll out the accessibility features themselves since we wouldn't have access to the handling of the `activeTabIndex`. We could try to mirror the state controlling the component based on which tab has a `selected` tab set to true, but maintaining two states to control the same component seems like an anti-pattern to me which could potentially open up bugs.

androche:
@meggettt well its working with tabs in the controlled component now so wondering if it might be easy enough to swap out the tab for arrow instead https://magna-react.vercel.app/components/tab

meggettt:
> @meggettt well its working with tabs in the controlled component now so wondering if it might be easy enough to swap out the tab for arrow instead https://magna-react.vercel.app/components/tab

Ok, I've added some functionality and did some refactoring so that the controlled components also have the same accessibility feature. They should work with the directional keys now.

rwharrisjr:
A few things

- can't seem to tab into the container for the controlled usage
- The "More" menu tab "button" should show as "active" when it's open
- focusing into the tab container doesn't give an indication of any active element - makes it hard to realize that you can arrow left/right
- Arrowing through a tab group with hrefs/links open the link when the tab is hit. In this case the tab element should focus, but not trigger the event handling.

meggettt:
> Visit Preview

Good points, but I would like to have a discussion about the "More" button being active when it's selected. I felt that it should only be active when one of it's menu items was selected, that way it's always clear which tab is active in relation to the content currently being displayed. Otherwise, currently if you selected the "More" tab then hit exit, we just lose an active tab altogether.

androche:
> > Visit Preview
> 
> @rwharrisjr Good points, but I would like to have a discussion about the "More" button being active when it's selected. I felt that it should only be active when one of it's menu items are selected; that way it's always clear which tab is active in relation to the content currently being displayed. Otherwise, currently, if you selected the "More" tab then hit exit, we just lose an active tab altogether.

hmm I had to take a second to think about it - I think comparing it with how a mouse user would interact with tabs would be best and then replicating that behavior. 

So a mouse user would:

1) Hover over the more button 
2) click it (making the button active)
3) dropdown opens and then user can make a selection.

For replicating that action with arrows 

1) Arrow over to the more button
2) Hit enter to select it (making button active)
3) Dropdown opens

meggettt:
> > > Visit Preview
> > 
> > 
> > @rwharrisjr Good points, but I would like to have a discussion about the "More" button being active when it's selected. I felt that it should only be active when one of it's menu items are selected; that way it's always clear which tab is active in relation to the content currently being displayed. Otherwise, currently, if you selected the "More" tab then hit exit, we just lose an active tab altogether.
> 
> hmm I had to take a second to think about it - I think comparing it with how a mouse user would interact with tabs would be best and then replicating that behavior.
> 
> So a mouse user would:
> 
> 1. Hover over the more button
> 2. click it (making the button active)
> 3. dropdown opens and then user can make a selection.
> 
> For replicating that action with arrows
> 
> 1. Arrow over to the more button
> 2. Hit enter to select it (making button active)
> 3. Dropdown opens

I understand the thought process, but I think that still leaves us with the original issue. As it stands, I believe the way the tabs work is kind of broken. So replicating it leads to the same issue I mentioned above.

androche:
@meggettt When in doubt its nice to check here!

https://magnetic.cisco.com/0a43ab5cd/p/89dccd-tab

I do see the dilemma now you are talking about now - **on hover** it shows a dropdown menu and then active once selected.  So to try to add to discussion on your initial question, it seems like keeping the automatic tigger you have on tabs would emulate that open on hover action, but if there's a way to prevent event triggers _just_ for links that would solve that issue.

rwharrisjr:
I see what y'all mean with the More button state. I'm fine leaving it as is. The menu opens as expected and that's really what matters.

rwharrisjr:
Closing in favor of #706 706

PR #655, opened 2024-03-21
Remove padding from p tags in AAlert

rwharrisjr:
The markdown pages on the demo site seem to add a `<p>` tag to the AAlert message content when used outside the Playground. This causes the alignment in alerts to be a bit off

### DIFF
--- a/framework/components/AAlert/AAlert.scss
+++ b/framework/components/AAlert/AAlert.scss
@@ -129,5 +129,5 @@ $alert-close-icon-size: 10.5px;
 }
 
 .a-app .a-alert p:first-child {
-  margin: 0;
+  margin: 0 0 ($spacing * 0.5) 0;
 }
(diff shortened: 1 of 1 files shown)


framework/components/AAlert/AAlert.scss
```diff
@@ -126,3 +127,7 @@ $alert-close-icon-size: 10.5px;
     }
   }
 }
+
+.a-app .a-alert p {
+  margin: 0;
+}
```
brennarvo, line 133: I would maybe consider adding this to the demo site specifically. This might cause some unexpected margin resets in other apps usage of `AAlert`, like below:

```jsx
<AAlert>
  <p>First paragraph</p>
  <p>Second paragraph</p>
  <p>Third paragraph</p>
</AAlert>
```
The `.a-app .a-alert p` will remove the margins between each paragraph

androche:
Could i have a sample - I'm not sure what to recreate to verify

rwharrisjr:
It's only inside magna

Screenshot 2024-03-26 at 8 47 54 AM

androche (approved):

PR #657, opened 2024-03-21
Fix double focus for checkboxes with no label

rwharrisjr:
When using without a label, the wrapper has a tabIndex for focus, along with the `<input>`. This causes two separate focus states when tabbing through.


androche:
Having trouble recreating this - also enabling checkbox on enter appear to no longer work!

Screenshot 2024-03-26 at 9 44 28 AM
Screenshot 2024-03-26 at 9 44 12 AM

androche:
Also meant to add that's an unrelated issue as it is seen on prod, but wanted to mention it!

androche (approved):

PR #659, opened 2024-03-21
Check if focus is in an input element before trapping arrow keys

### LINKED ISSUE #658: Cursor does not move when using right arrow key when `ATextInput` is nested inside of `AListItem`
**Describe the bug**

When `ATextInput` is nested inside of a `AListItem`, the right arrow key to move the cursor stops working.

**To Reproduce**

1. Go to a Playground editor on the docs site, e.g. https://magna-react.vercel.app/components/list#usage
2. Paste the following into the code playground:
    ```
    <AList>
      <AListItem>
        <ATextInput />
      </AListItem>
    </AList>
    ```
3. Type "Hello world" into the text input
4. Press your left arrow key
5. Press your right arrow key
6. **Observe that the cursor stays put**

**Expected behavior**

The cursor should move back to the original position when pressing the right arrow key.

 - Device: Desktop
 - OS: macOS
 - Browser Chrome
 - Version 122.0.6261.94

rwharrisjr:
Resolves #658

### DIFF
--- a/framework/components/AMenu/AMenu.js
+++ b/framework/components/AMenu/AMenu.js
@@ -94,9 +94,15 @@ const AMenu = forwardRef(
         e.stopPropagation();
         const previous = getPrevious();
         previous && previous.focus();
-      } else if (e.keyCode === keyCodes.right) {
+      } else if (
+        e.keyCode === keyCodes.right &&
+        document.activeElement.tagName !== "INPUT"
+      ) {
         e.preventDefault();
-      } else if (e.keyCode === keyCodes.left) {
+      } else if (
+        e.keyCode === keyCodes.left &&
+        document.activeElement.tagName !== "INPUT"
+      ) {
         e.preventDefault();
         submenu && closeHandler(e);
       } else if (e.keyCode === keyCodes.down) {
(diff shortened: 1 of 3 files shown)


framework/components/AMenu/AMenu.js
```diff
@@ -94,9 +94,15 @@ const AMenu = forwardRef(
         e.stopPropagation();
         const previous = getPrevious();
         previous && previous.focus();
-      } else if (e.keyCode === keyCodes.right) {
+      } else if (
+        e.keyCode === keyCodes.right &&
+        document.activeElement.tagName !== "INPUT"
```
brennarvo, line 99: How would you feel about linking the issue that inspired this condition as a comment above it? Not required, but might add some context since arrow key logic is always finicky

brennarvo: Actually, that's what git is for.

brennarvo: Okay — last comment haha. Did you also mean to add this check for the `tagName` in the `AListItem` component, too?

https://github.com/cisco-sbg-ui/magna-react/blob/5d1cb8877c793acf49bcb5d86cade4b873d090ab/framework/components/AList/AListItem.js#L81-L99

brennarvo (approved):

androche (approved):
Pulled and tested

PR #661, opened 2024-03-22
MultiSelect readOnlyInput, fixes

rwharrisjr:
Adds

- `readOnlyInput` prop to leave menu active, but set the text input to `readOnly`
- `filterFunction` example in docs

Fixes

- `readOnly` state background color in inputs


jflook10 (approved):
Tested in vercel preview, works as expected 👍

PR #665, opened 2024-03-25
Support classic themes; setup for future

### LINKED ISSUE #630: Support all available Magnetic themes
My application is using the "classic" variants of the Magnetic themes, and uses the design tokens provided by the CDS library. I noticed that the magna-react root element only ever uses the default Magnetic themes, which will result in mismatched colors for things like status.
image

Support in magna-react for all available Magnetic themes. 

Also, we'd like for the internal design tokens used by magna-react to not take precedence over local overrides. We have `brand-accent` overrides in our application to achieve a blue product theme (instead of the default green Magnetic theme) and with the latest version of magna-react, these overrides are not taking affect.

rwharrisjr:
Adds support for the magnetic "classic" (classic light and classic dark) themes.

Treats themes as two groups, light themes and dark themes. This should continue to support use of the `AThemeContext`'s props `isLight` and `isDark` as before.

Additionally, the theme constants are now exported for use in other libraries.

Resolves #630

Screenshot 2024-03-25 at 11 47 03 AM

### DIFF
--- a/docs/Sidebar.js
+++ b/docs/Sidebar.js
@@ -33,13 +33,15 @@ const GitHubIcon = (props) => {
 const Sidebar = ({menus, currentDoc}) => {
   const [items, setItems] = useState([]);
   const currentDocRef = useRef();
-  const {currentTheme} = useATheme();
-  const styleColor = currentTheme === "dusk" ? "white--text" : "black--text";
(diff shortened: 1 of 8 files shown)


androche (approved):
NICE

PR #667, opened 2024-03-26
Safety fallback useFocusTrap when node is no longer in the DOM

rwharrisjr:
In cases such as virtualized lists that update dynamically, the next node from the TreeWalker may no longer exist in the DOM. The memory reference still points to the object, and since the reference holds onto it the object isn't nulled out. However, the node may be detached. In this case the `tab` key stops working, since there is no previous or next node for it to move to. 

As a safety, when the next node to focus is no longer in the DOM, fallback to refocus to `rootRef` element.


androche (approved):
👏

PR #672, opened 2024-04-01
Reapply modal accessibility enhancements

rwharrisjr:
Reapply the earlier commits for moving forward with v2.x


androche (approved):
Tested in deployment

PR #674, opened 2024-04-04
Doc redesign take2

rwharrisjr:
This does a redesign and standardization of doc pages, with enhancements for developer quality of life (implemented by @brennarvo)

Also makes things look more magnetic-y


brennarvo:
I still can't believe you managed to sift through my messy POC code. *Thank you*. I never thought I'd see the day this was ever finished 😂

One bit of feedback — should we also add the `max-width` to the main content area, kind of like it is in the headers? Kind of like how in this outdated POC, the content under the header also has the same `max-width`: https://magna-react-beta.vercel.app/

Screenshot 2024-04-04 at 10 42 30 AM

Actually, I have a feeling you did this on purpose so that the doc pages had more room to work with (since the code snippets can be long)

rwharrisjr:
I think the pages should be wide, but I did fix the width restrictions on the header

androche (approved):
@rwharrisjr @brennarvo This is flawless - Out of all of this I could only comment on some sidebar spacing/colors. 😄

PR #687, opened 2024-04-16
add deprecations/migration to changelog; fix component props drawer

rwharrisjr:


androche (approved):
Screenshot 2024-04-16 at 3 10 23 PM

Can we make the content a lighter background?  It's kind of hard on the eyes and would help to differentiate from sidebar

androche (approved):

PR #696, opened 2024-04-30
Fix test failure - give secondary element for test to tab to 

rwharrisjr:
Fixes the ATextInput failure by giving a tababble element for `cy.tab()` to move into


androche (commented):
Weird - havent looked at the test, but could we just delete the extra cy.tab()?

androche:
Maybe swap to `cy.get(".a-text-input__input").blur();` to test rather than the extra element?

androche (approved):

PR #700, opened 2024-05-06
fix proptypes and map key for AActivityTimeline

rwharrisjr:


androche (approved):

androche:
Actually this is somehow moved over a bit.  This is a little different when i viewed locally first time around: Can we bump it over

Screenshot 2024-05-06 at 4 48 39 PM

PR #706, opened 2024-05-07
Refactor tab selected tracking; add arrow key accessibility

rwharrisjr:
Builds off the ideas in https://github.com/cisco-sbg-ui/magna-react/pull/650

The previous implementation lacked a way to track focus and selected state simultaneously. This made it difficult to configure the keyboard arrows to not automatically select a tab (not desired when the panel content may take time to load). Additionally it made it difficult to avoid opening link tabs in the auto-select mode while tracking the other states.

### DIFF
--- a/framework/components/ATabs/ATabGroup.js
+++ b/framework/components/ATabs/ATabGroup.js
@@ -13,6 +13,38 @@ import ATabContext from "./ATabContext";
 import {useCombinedRefs} from "../../utils/hooks";
 import "./ATabs.scss";
 
+const getNextFocusId = (direction, containerEl, focusedEl) => {
+  let nextId, toFocus;
+
+  if (direction == "ArrowLeft") {
+    toFocus = focusedEl.previousSibling;
+
+    while (!toFocus || toFocus?.classList.contains("hide")) {
+      if (!toFocus) {
+        toFocus = containerEl.lastChild;
+      } else {
+        toFocus = toFocus.previousSibling;
+      }
+    }
+
+    nextId = toFocus.getAttribute("data-tabid");
+  } else if (direction == "ArrowRight") {
+    toFocus = focusedEl.nextSibling;
+
+    while (!toFocus || toFocus?.classList?.contains("hide")) {
+      if (!toFocus) {
+        toFocus = containerEl.firstChild;
+      } else {
+        toFocus = toFocus.nextSibling;
+      }
+    }
+
+    nextId = toFocus.getAttribute("data-tabid");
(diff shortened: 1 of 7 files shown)


framework/components/ATabs/ATabGroup.js
```diff
@@ -144,17 +207,54 @@ const ATabGroup = forwardRef(
             </AListItem>
           );
         }
+
         return <AListItem {...rest} />;
       }
     });
 
     return (
-      <div {...rest} role="tablist" ref={combinedRef} className={className}>
```
androche, line 152: There's still no way for someone on a keyboard to get to the tabbed section.  In ARIA docs I see that it immediately goes to the tab (showing the container content). But I dont think that would work for most use cases since there will be competing focuses happening on a page.

If there's a way to make the TabGroup have a container focus accessible through the tab key or at least allow tab (key) into the first tab, that might help.

framework/components/ATabs/OverflowMenuTab.js
```diff
@@ -53,21 +54,35 @@ const OverflowMenuTab = ({children}) => {
...
+            tabGroupRef.current?.blur();
+            menuRef.current?.focus();
+          } else {
+            tabGroupRef.current?.focus();
+          }
+        }}
       >
         More <AIcon>{menuIcon}</AIcon>
       </ATab>
       <AMenu
```
androche, line 74: When leaving the menu to select another tab, the menu doesnt close and it doesnt allow you to select a new tab
Screenshot 2024-05-10 at 9 36 56 AM

androche: uh oh - vertical 
Screenshot 2024-05-10 at 9 38 25 AM

androche: Something weird with links - 

links

androche (commented):
A few initial comments upon testing.  These tabs are ridiculously tricky, thanks for taking this on, there was no way I could do it again 😆   I can confirm dynamic more button is still working!  There is an existing bug where it sometimes will overflow on initial load when tabs are already in a collapsed state.

rwharrisjr:
@meggettt fyi, I took a lot of your logic for this PR. To make the `selectOnArrow` prop work I realized the original code needed some refactoring, so ran with that and built the arrow left/right functionality off your code.

androche (commented):
That seems to get all of it! 🎉  One comment below, other than that its ready

androche (approved):

PR #716, opened 2024-05-10
Update breakpoints to magnetic spec; move code into hooks

### LINKED ISSUE #671: Add Magnetic breakpoints
### **Is your feature request related to a problem? Please describe.**

Yes, we are using the Magnetic design system, which has different breakpoints than are what's in magna-react. There are a lot of discussions happening about design changes across XDR that are supposed to be using the Magnetic defined breakpoints, so we'll need those in this library.

Also, I could not find what the magna-react breakpoints were anywhere on the [docs site](https://magna-react.vercel.app/services/breakpoint). It would be helpful if all of the breakpoints were listed somewhere.

**Edit:** I found the breakpoint definitions listed on the [helper classes page](https://magna-react.vercel.app/components/app#helper-classes). It might be helpful to put it somewhere more prominent or at least link to this page from `useMediaQuery` and `useABreakpoint`.

Magna-react breakpoints are:

xs: 479px
sm: 767px
md: 992px
lg: 1200px
xl: 1440px

### **Describe the solution you'd like**

Add Magnetic breakpoints to magna-react. I know we can't rip the old ones out or change them for backward compatibility, so maybe the Magnetic breakpoints could be added in addition?

Also, is there a way to expose these variables for us to use in our app similar to color tokens?

[Magnetic breakpoints](https://magnetic.cisco.com/0a43ab5cd/p/077cd1-layout--page-templates/t/57e483)

Screenshot 2024-03-29 at 1 35 46 PM

rwharrisjr:
Resolves #671 

This updates the breakpoints to magnetic spec, and exports them as an object for use outside of the hook. Docs have been updated as well under `/hooks/use-a-breakpoint`

@sabrinamokerji fyi

This is a breaking change, so will go into magna 2.0

### DIFF
--- a/framework/components/AApp/AApp.mdx
+++ b/framework/components/AApp/AApp.mdx
@@ -511,11 +511,12 @@ For example, `d-flex d-md-block d-print-none`:
 
 Media queries are set for the following breakpoints:
 
-- xs: 479px
-- sm: 767px
-- md: 992px
-- lg: 1200px
-- xl: 1440px
+- xs: 420
+- sm: 960
+- md: 1280
+- lg: 1680
+- xl: 2080
+
 - print (for print styles)
 
 ### Reference
--- a/framework/changelog.mdx
+++ b/framework/changelog.mdx
@@ -19,6 +19,18 @@ useContainer: true
 - `AMenu` now automatically has `useFocusTrap` to support tabbing inside menus after opening
 - `ATabGroup` now adheres to ARIA spec in regards to tabs and left/right arrow keys
 
+### Magnetic Changes
+
+#### Breakpoints
+
+Breakpoints have been updated to match magnetic spec.
+
+- xs: 420
+- sm: 960
+- md: 1280
+- lg: 1680
+- xl: 2080
+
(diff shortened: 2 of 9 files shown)


framework/components/AApp/AApp.mdx
```diff
@@ -511,11 +511,12 @@ For example, `d-flex d-md-block d-print-none`:
 
 Media queries are set for the following breakpoints:
 
-- xs: 479px
-- sm: 767px
-- md: 992px
-- lg: 1200px
-- xl: 1440px
+- xs: 420
+- sm: 960
```
sabrinamokerji, line 515: @rwharrisjr does Magnetic really not have a breakpoint between 960-420? There is usually always one in the 700s

rwharrisjr: I looked and didn't see anything. They actually don't even have the xs 420px - so I just guessed at that. But we can add more in-between with whatever works best for us

androche (approved):
I didn't really test this 🙃 

Let me know if there's something specific or if its just a see something say something as we all go.

PR #717, opened 2024-05-15
feat: resurrect tertiary no padding

androche:
I know we talked about a new component, but wanted to try this out since it got in once before. Let's test around and see how this works.

### DIFF
--- a/framework/styles/utilities/mixins.scss
+++ b/framework/styles/utilities/mixins.scss
@@ -113,6 +113,11 @@
 // -----------------------------------------------------------------------------
 // Button sizes
 // -----------------------------------------------------------------------------
+
+.remove-padding {
+  padding: 0;
+  margin: 0 12px;
+}
 @mixin medium-button {
   height: 34px;
   padding: 8px 12px;
@@ -122,6 +127,11 @@
     svg.a-icon {
       width: 16px;
     }
+  }
+
+  &.a-button--tertiary {
+    height: 20px;
+    @extend .remove-padding;
   }
 }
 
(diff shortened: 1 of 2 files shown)


androche:
Gave this a shot and it did impact several implementations in Ribbon where tertiary is used as a button (not a link).
Will try with suggestion of inline prop next

framework/styles/utilities/mixins.scss
```diff
@@ -113,6 +113,11 @@
 // -----------------------------------------------------------------------------
 // Button sizes
 // -----------------------------------------------------------------------------
+
+.remove-padding {
```
rwharrisjr, line 117: probably should namespace this `.a-remove-padding`

framework/components/AButton/AButton.cy.js
```diff
@@ -26,4 +26,14 @@ describe("<AButton/>", () => {
...
   });
+
+  it("should have no padding", () => {
+    cy.mount(
+      <AButton noPadding tertiary>
+        No padding
+      </AButton>
+    );
+
+    cy.get(".a-button").should("have.class", "a-button--inline-btn");
```
rwharrisjr, line 37: I think to be thorough you could also test

`cy.get(".a-button").should('have.css', 'padding', '0px')`

rwharrisjr (changes requested):

framework/components/AButton/AButton.cy.js
androche: Good call

PR #723, opened 2024-05-16
Revert "fix: accordian transition"

rwharrisjr:
This reverts commit d50888d24320ded92e2fc772930b61450e11d1ed

https://github.com/cisco-sbg-ui/magna-react/commit/d50888d24320ded92e2fc772930b61450e11d1ed

`overflow: hidden` seems to break ASelect menu when inside of an accordion

Reverting for now


androche (approved):
Will follow up because I think the original issue might be able to be fixed without a magna change

PR #724, opened 2024-05-16
feat(AActivityTimeline): allow timeline item connector visibility to be controlled with prop

brennarvo:
## Context

When timeline items are rendered from a paginated source of data, it is possible (and likely) that the very last item in the user's current view is *not* the very last item in the source of data. In such cases, the vertical "connector" rendered underneath the timeline *should* be rendered. However, given the current implementation, the connector would be hidden since it is the last list item in the DOM.

This PR solves this issue by allowing a developer to pass a `withConnector` prop to control the connector's visibility.

(Please see [preview branch "Paginated List" example](https://magna-react-git-feat-timeline-item-connector-prop.securex-preview.app/components/activity-timeline) for more context)

## This PR

* Adds a `withConnector` prop to the `AActivityTimelineItem` component
* Adds documentation for paginated lists of data with the timeline item
* Adds and refines tests

## Test Plan

1. Navigate to the [`AActivityTimeline` documentation page for the preview branch](https://magna-react-git-feat-timeline-item-connector-prop.securex-preview.app/components/activity-timeline)
2. In the very first example, pass the `withConnector` prop to the very last item that is rendered in the timeline
3. **Verify that the timeline item connector renders**

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -111,3 +111,12 @@ Cypress.Commands.add("hoverATooltip", (withClick = false) => {
 Cypress.Commands.add("clickATooltip", (withClick = false) => {
   return cy.getByAriaLabel("information icon").trigger("click");
 });
+
+Cypress.Commands.add("getPseudoElementStyle", ($el, pseudoElement, style) => {
+  // https://stackoverflow.com/a/75887385
+  const subject = $el[0];
+  const window = subject.ownerDocument.defaultView;
+  const pseudoElementStyles = window.getComputedStyle(subject, pseudoElement);
+
+  return pseudoElementStyles.getPropertyValue(style);
+});
--- a/framework/components/AActivityTimeline/AActivityTimeline.cy.js
+++ b/framework/components/AActivityTimeline/AActivityTimeline.cy.js
@@ -19,13 +19,13 @@ describe("<AActivityTimeline />", () => {
       if (index < $list.length - 1) {
         expect(borderValue).contains(DEFAULT_BORDER_STYLE);
       } else {
-        // The very last item should not have a left border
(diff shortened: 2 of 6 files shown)


androche:
@brennarvo  I haven't worked in the new timeline yet and apologies if you've already considered this - just curious would/could this be a case for something like "isLastItem" in the list and then add class of no connector inherently?  That way it wouldn't depend on css lastchild in the current view.

androche:
> @brennarvo I haven't worked in the new timeline yet and apologies if you've already considered this - just curious would/could this be a case for something like "isLastItem" in the list and then add class of no connector inherently? That way it wouldn't depend on css lastchild in the current view.

Ah please disregard - I think i'm following now.  I see its not just a scroll issue, but literally the items aren't there to be found in DOM.  Makes sense!

brennarvo:
@androche it looks like you answered your own question 😄

I actually did have it as `isLastItem` at first, but then I figured there might be a case in the future where the connector should stop before the last item anyways, like this:

Screenshot 2024-05-16 at 3 46 46 PM

Then, `isLastItem` gets a little confusing since that style in the screenshot above would be applied to something that isn't the last item

brennarvo:
@androche also — apologies for the last minute push. I found a refactoring opportunity in the tests. I had copied/pasted the same `getComptedStyle` logic enough that I figured it should be its own Cypress command

androche (approved):
I think this all looks/works great!!  The only thing I stumbled on was when I typed in 40 into the example, there were no connectors on the last page, but 40 isn't divisible by 3. :) Just pointing it out, but I don't think that state would happen during actual implementation.

Screenshot 2024-05-16 at 5 12 54 PM

PR #739, opened 2024-05-23
proxy click router component as tab; don't style links at a global level for tabs

rwharrisjr:
When the ATab is a router element, the onClick functionality isn't actually what controls the route change - so proxy click the item in the more menu to run the router's logic.

Color causes issues with tabs that set an `<a>` element as the component for a tab

### DIFF
--- a/framework/components/ATabs/ATabGroup.js
+++ b/framework/components/ATabs/ATabGroup.js
@@ -115,7 +115,7 @@ const ATabGroup = forwardRef(
         parseInt(tabStyle.marginLeft) + parseInt(tabStyle.marginRight);
 
       //Calculate width of overflow tab which serves as our minimum width.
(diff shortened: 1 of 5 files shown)


androche (approved):

framework/components/ATabs/ATabGroup.js
```diff
@@ -89,80 +91,78 @@ const ATabGroup = forwardRef(
...
 
-      const target = combinedRef.current.parentNode;
-      const resizeObserver = new ResizeObserver(() => {
-        if (combinedRef.current) {
-          handleOverflow();
-        }
-      });
+      const target = combinedRef.current;
+      const resizeObserver = new ResizeObserver(
+        debounce((foo) => {
```
androche, line 160: foo

androche (approved):
Pulled and retested - works well in IM!!  One typo, but approving ahead of time.

androche:
Actually wait! I tested without the "improvements" commit

androche (approved):

Issue #743, opened 2024-05-31
AAccordion is still occupying space (height) when collapsed


bredy:
**Describe the bug**

When I have `AAccordion` component with longer bodies and I collapse it to headers only the height of `AAccordion` still matches uncollapsed state.

**To Reproduce**
- Create `AAccordion` with multiple bodies with longer content
- Collapse all headers
- Check in inspector height of `AAccordion` element

**Expected behavior**
- height matches just collapsed headers

**Screenshots**
- height exceeds headers and matches height of uncollapsed state

**Additional context**

This is issue in "incident-manager" caused by this issue :
![Incident Report - Scroll Bar](https://github.com/cisco-sbg-ui/magna-react/assets/178547/378fb445-79ae-4ea9-9685-e730510e9f9a)

Scrollbar reflects the actual height, for collapsed state there should be no scrollbar but there is a huge one.

PR #744, opened 2024-05-31
fix: AAccordionBody content hidden in collapsed state

bredy:
- height of element is now reduced to visible part when it is in collapsed state

Closes: #743

Now the `AAccordion` when collapse has height related to its headers. Body content does not enlarge element height now.
In "incident-manager" it fixes two places when `AAccordion` is used - "report tab" and "response tab".

Yes, this fixed behaviour could be breaking change.


rwharrisjr (approved):
Seems to work, and doesn't break the animation

androche:
@rwharrisjr we had to revert for Vanya's dropdown issue.  Is there a way we can test that?

androche:
Verified that this change does not introduce the same regression.  

Screenshot 2024-06-05 at 12 45 24 PM

rwharrisjr (approved):

PR #745, opened 2024-05-31
change to node 18.19 to workaround node issue on the action runner

rwharrisjr:
Seems there's an issue with the node version on the action runner causing npm ci to fail. This works around it for now


bredy (approved):
LGTM

PR #764, opened 2024-06-25
Add magnetic compliant pagination component

### LINKED ISSUE #763: Align pagination page size selection with design
Rather than an input it should be a dropdown

- Rows displayed per page: 10, 20, 30 (default), 50, and 100
- Text should read `Rows per page` *before* the dropdown input

rwharrisjr:
Resolves #763 

Screenshot 2024-06-25 at 1 52 48 PM

The old pagination component doesn't align well to magnetic in display, page counts, or style. It also has variants that are not part of the magnetic design.

This PR adds a new `APaginator` component that aligns with the magnetic design. `APagination` should be considered deprecated.


androche (commented):
Wow that was fast!!!  Will look soon - At a quick glance it seem there's some minor differences in figma that could be quickly updated like width of input seems to be 64px, height of buttons 28px, and can remove the 1-15 text altogether.  
Screenshot 2024-06-28 at 9 57 10 AM


I'll pull up the other magnetic update documentation as well.

framework/components/APaginator/APaginator.scss
```diff
@@ -0,0 +1,95 @@
...
+    margin: 0;
+    .a-icon {
+      fill: var(--control-border-default);
+    }
+  }
+
+  .a-button,
+  .a-button--selected {
+    min-width: 32px;
+    height: 32px;
```
androche, line 20: Height and width should be `28px`

framework/components/APaginator/APaginator.scss
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  &--disabled {
+    pointer-events: none;
+    > .a-button-group {
+      > .a-button-group__wrapper {
+        > .a-button {
+          color: var(--base-text-weak-default);
+          &--selected {
+            border: 2px solid var(--interact-border-disabled) !important;
```
androche, line 50: control-border-medium-default

androche (changes requested):
Update: I see now actually figma leaves out the 1-15 from the magnetic specs so disregard.  Here's a few more specific items from the figma.

framework/components/APaginator/APaginator.scss
```diff
@@ -0,0 +1,95 @@
...
+  display: flex;
+  align-items: center;
+
+  &__previous,
+  &__next,
+  &__first,
+  &__last {
+    margin: 0;
+    .a-icon {
+      fill: var(--control-border-default);
```
androche, line 13: interact-icon-weak-default

framework/components/APaginator/APaginator.scss
```diff
@@ -0,0 +1,95 @@
+@import "../../styles";
+
```
androche, line 2: Not in figma, but needs a hover state - probably same as the selected background

Screenshot 2024-06-28 at 10 33 21 AM

Screenshot 2024-06-28 at 10 33 34 AM

rwharrisjr:
@androche the width of the ASelect was intentional, any smaller and when 100 is selected the width changes and the controls move. Decided to avoid that and be slightly different than the figma

androche:
> @androche the width of the ASelect was intentional, any smaller and when 100 is selected the width changes and the controls move. Decided to avoid that and be slightly different than the figma

Hmm it looks like that would be more related to the padding on `.a-select__selection`

So if you take away the padding on right pushing the icon and similarly left on icon it prevents the jumping
```scss
 .a-select {
      width: 64px;
      .a-icon {
        padding: 0 5px;
      }
      &__selection {
        padding: 0 0 0 8px;
      }
    }
 ```
 
jumping

androche:
Added fixes to the button styling and colors:

Screenshot 2024-07-01 at 9 30 05 AM

This is almost there, but it looks like disabled is breaking in this PR.  Currently, implementing disabled should disabled all buttons and controls as seen below. 

Screenshot 2024-07-01 at 9 49 34 AM

rwharrisjr:
Fixed the disabled state

androche (approved):
🎉  Looks/working great! We may want to move over the props to this one (or link it), but we could do that when we remove the old one.

androche (approved):

Issue #771, opened 2024-06-28
Add Typescript and Eslint


androche:
Issue for reference to ongoing efforts

androche:
Closing as majority of work has been done.  All that is remaining is converting our internal files.

PR #773, opened 2024-07-02
Magna Typescript and Eslint

androche:
#771 **Adds Typescript and Eslint**

**Typescript**
Non-breaking which means we can add to on an ongoing basis and types will become available for component use as they get updated.

**Eslint**
Incorporated and fixed issues


framework/components/ATriggerTooltip/ATriggerTooltip.cy.js
```diff
@@ -131,119 +131,118 @@ describe("<ATriggerTooltip />", () => {
...
-        content={<span data-testid="tooltip-content">test content</span>}
-        {...triggerTooltipProps}
-      >
-        <span>test anchor</span>
-      </ATriggerTooltip>
-    </div>
-  );
-}
+// function CustomTriggerTest(triggerTooltipProps) {
+//   const btnRef = useRef(); //TODO this is not updating in time for the test to work
```
androche, line 233: Needs debugging! Cypress only issue: `useRef` is now undefined and passes no ref to button or trigger tooltip

brennarvo:
@androche this is exciting to see. If you want some point of reference, the [incident manager repo also defined types for all the Magna React components](https://github.com/advthreat/incident-manager/tree/master/src/types/magna-react). They're not necessarily the most up to date, but they could be useful when trying to incorporate the `component` and `as` props with TypeScript in the Magna React repo.

androche:
Thanks @brennarvo! :) Hoping we can get this going and then chip away over time.  Yes the files in IM have been extremely helpful!!  After this 1st pass, hoping to spread the word around to get some folks to double check these general types to make sure we'll be heading in the right direction for remaining components.

rwharrisjr (approved):
nice!

androche:
@mvelkCisco @lukaskral @brennarvo @OlgaBilogurova, @bredy 

If any of you all wouldn’t mind taking a very brief look at a couple (doesn't have to be all) of these typescript starter components to make sure the types are lining up with IM (particularly around various refs) that would help us set a good foundation for these patterns for the remaining components.

`AAccordion`, `AIcon`, `ATriggerTooltip`, `ATooltip`, `AMenuBase`, `AAppContext`

Please ignore the 90 file changes! - A good portion of this PR also includes eslint fixes.

mvelkCisco:
Scanning this quickly, it looks nice! Thank you for doing this.

You probably know that incident manager has a lot of magna react types already: https://github.com/advthreat/incident-manager/tree/master/src/types/magna-react. They are not perfect, but could be used as a starting point, or for cross-checking.

If I remember correctly, the hardest thing to figure were polymorphic components (the ones that accept other component to render in the `as` or `component` prop). The issue is that we want to inherit the props from the `as` component, when that prop is set, and fallback to default otherwise. We have a generic utility for that, eg used here: https://github.com/advthreat/incident-manager/blob/master/src/types/magna-react/ADrawer.js.ts It might look a bit scary and a bit tricky to use/debug sometimes, but it should work.

We've also added an `Override` utility you will see used everywhere. The purpose of it is to replace the types of fields, instead of creating unions. For example for `type TypeA = {a: string, x: string}` and `type TypeB = {a: number, y: string}`, the result of a simple intersection `TypeA & TypeB` is `{a: string | number,  x: string, y: string}`, but we usually don't want that union `a: string | number` type. `Override` omits the original filed types, so the result should be `{a: number,  x: string, y: string}`. 

Another thing to keep in mind is that a lot of magna-react components spread the "rest" of the props to a particular html element and we wanted to support that html element's props too. Eg. in case of checkbox, the base element is div https://github.com/advthreat/incident-manager/blob/720d9b0b9bdb543c8a4335342e57b1bdd8e0103c/src/types/magna-react/ACheckbox.js.ts#L12. 

Let me know if you have any questions or need help.

androche:
Thank very much @mvelkCisco!!  This will be extremely helpful to reference especially when we get to those `as` components.  I think we will need to think more about/wrap head around the overrides as I did leave them out in at least one or two of these components here.  Also thanks for clarification on the `componentPropWithRef` as I was most worried about the refs.  @rwharrisjr and I will study a bit on this and adjust before the next series!

androche:
@mvelkCisco I noticed that `div` is used the most so I was thinking to cut down on repeating we could add the withRef to the rest of our most reused types that we're extending to interfaces. We'd have to still the write in for other elements which would be ok.

ex:
```
export type BasicComponentProps = React.ComponentPropsWithRef<"div"> & {
  children?: React.ReactNode;
  className?: string;
};
```

I also see the override.d.ts in IM, but I think we're still trying to understand how that can help magna or if its something specific to IM?  If you could advise a little more on that and if you think it's important that we continue that pattern over here, that would be great!

mvelkCisco:
`React.ComponentPropsWithRef<"div">` vs `BasicComponentProps`: You can do that, but I'm not sure if it would cut down on the repetition all that much. You still need to use one or the other every time, and I feel like seeing a `ComponentPropsWithRef<"div">` is more explicit and clear, at least to me.

Regarding the specific `BasicComponentProps ` example: `className` prop should not be needed. `className` should already be included in `ComponentPropsWithRef<"div">` (it's a property of react `div` element). Same with the `children`. I see `className` added here and there in incident manager types, but in most cases it seems redundant. I'm not sure what other props I would include in this "BasicComponentProps" type.

Also this sets you up on a path, where the first time you have a component that should not have one of these 3 things (`className`, `children` or `ref`), someone will make a mistake and just use `BasicComponentProps` anyway. To be fair, it could happen with the `ComponentPropsWithRef<"div">` too, it's just a bit more obvious what's happening.

`override.d.ts`: This is tricky. Let me explain further. 

It's needed because of a combination of TS behavior of "intersection" in general (`TypeA & TypeB`), with the fact that magna-react (and also incident manager) sometimes re-define some props with a different type when extending/composing components, and then often just spread the rest of the props to the original component. This is kind of hard to reflect in types (and probably would not be implemented in this way if this was done in TS from the start).

I actually made a mistake in my previous comment. Regardless, [see this example](https://www.typescriptlang.org/play/?ts=5.5.3#code/C4TwDgpgBAKuEEEoF4oG8CGAuKBnYATgJYB2A5gNxQBGO+x5ANFAB52GlkC+VA9L1AAWGXFABEGMVABmRADYQAJgChQkWPABCKdNigkArgFtqEAlVp4OTKCHYNufAcNESpsiHMVQA7kWCCUBhQikTS0mYQJMBQahDKqvAakADCOnCQSABkyRCaFAkAxgD2JPixEPg4GRBpqJg4AIzMlmLUYsxs4iwdtjhiIGI8UPyxgkSukjJEnt5+cnL6xTE+xQQA1szzi6ZQxdKxSWIkEABuZmIAhEA):

image

In my previous comment I've claimed `a` would be `string | number`, sorry about that. TS will actually make the `a` field of type `never`, so this object can't be constructed. At least in this simple example. 

This gets confusing fast, when we start using prop names that clash with the properties of the original element (or component) we are extending (and spreading the rest to) in magna react.

When I change the `Override` utility to just `T1 &  T2 &  T3 &  T4 &  T5`, I start getting errors like these:

image

In this case `ABadge` expects/allows `content` prop to be `ReactNode`:

image

But the `ABadge` is based on `ComponentPropsWithRef<"div">` (because that's where it spreads the rest of the props). And the `content` prop is supposed to be `string | undefined` in the "div" elements:

image

TS has issues reconciling this conflict. That's there `Override` comes in. 

```ts
type Override2<T1, T2> = Omit<T1, keyof T2> & T2;
``` 

literally means: first omit all the keys from T1 that are also in T2, then merge that with T2.

`Override2`, `Override3`, `Override4` etc. only exist because I could not quickly find a way how to write this for more than 2 types. The fields on left are always replaced by the fields on the right.

This situation only happens sometimes, when there is a prop naming clash. It's usually clashes with the html element types (and properties like `title`, `content`, I think I've seen `value`). Magna react also has a lot of "inheritance" (eg. `AFieldBase` -> `AInputBase` -> `ATextInput`), which might surface the same issue sooner or later (but I don't see any specific example right now).

To save us the trouble of identifying each case separately (errors can be a bit cryptic and don't show up until the prop is used), we just use the `Override` utility everywhere. It also makes everything nice and consistent, even though it's probably unnecessary in most cases.

This might have a more elegant solution, this is just something we came up with.

androche:
Wow thanks for this overview @mvelkCisco!    Re: `withRef` that was exactly my issue initially, since I wasn't including `ComponentPropsWithRef<"div">`, it was flagging className and children so it will be good to clean that up as I didn't know those were included!  Also thanks for taking the time to explain `Override` more - it would have been easy to copy paste it all in, but wanted to make sure I'd be able to explain it to someone as well! 🙂

rwharrisjr (approved):

PR #780, opened 2024-07-15
#779 - re-render to compute new size if multiselect input changes

### LINKED ISSUE #779: [BUG]: Issues with AMultiSelect
**Describe the bug**
Context: Issues observed during development of JIRA Q4 Epic
[XDR Notifications- Enable users to send notifications to instant messaging systems.](https://cisco-sbg.atlassian.net/browse/XDR-1561)

1.No placeholder is showing up, so user don't know where to type.
image

2. Label is not greyed out when component is disabled.
image

3. Seeing some UI Overlapping after adding initial search hit (screenshot)
Screenshot 2024-07-12 at 12 27 52 PM

**To Reproduce**

**Expected behavior**

**Screenshots**

 - Device: [e.g. Desktop, iPhone6]
 - OS: [e.g. iOS, Ubuntu]
 - Browser [e.g. Chrome, Safari]

**Additional context**

rwharrisjr:
Resolves #779 

When the input size changes due to tags, the render happens and computes the menu placement with the previous values.

This adds a resize observer to trigger a re-render *after* the size has actually changed

@wlorand3 @guptar8jan


androche (changes requested):
Looks like this is actually a regression from the original counter to tags refactor - it looks like when removing the onClick handler from `itemProps` line 426, the menu does not detect a change.  That's one area to look at, but if that doesn't work, wondering if we could utilize the `useResizeObserver` hook.

When I plugged it in by itself `useResizeObserver(multiselectRef)` it worked even without using the width.  We probably wouldn't want it to run on even instance, so maybe the ref could be on the tag multiselect and then return the observer if the ref does not exist.

rwharrisjr:
I tried one attempt with the onClick, but the size passed to AMenuBase was before the input size had actually changed due to the render order. I'll try with the hook

rwharrisjr:
Since the box shouldn't be changing size unless there are tags, I don't think the hook will be triggering re-renders unnecessarily now.

androche (approved):
Yes good point, and in fact if it were to ever change size we would want it to!

framework/components/AMultiSelect/AMultiSelect.js
```diff
@@ -70,6 +70,7 @@ const AMultiSelect = forwardRef(
     const [hasScroll, setHasScroll] = useState(false);
     const [filterValue, setFilterValue] = useState("");
     const [error, setError] = useState("");
+    const [, setUpdate] = useState();
```
androche, line 73: Can delete

androche (changes requested):

androche (approved):

PR #783, opened 2024-07-16
Remove margin right for spinner inside button

rwharrisjr:
Should be centered inside the button container

Screenshot 2024-07-16 at 11 32 49 AM


androche:
Looks like that was in there for the loader with text. 
Screenshot 2024-07-18 at 1 05 52 PM

Maybe a :not("") for something would do it

androche (approved):

PR #785, opened 2024-07-16
refactor: initiate typescript integration with cypress

brennarvo:
## Context

Along with slowly migrating components to TypeScript, we can do the same with unit testing. This will benefit the migration process since we ourselves can see how TypeScript for the components is working when we convert the tests to TypeScript.

## This PR

* Converts the `cypress` directory to TypeScript
* Creates a separate `tsconfig.json` file for Cypress Component Testing
* Converts the `useFocusTrap.js` file to TypeScript, along with its test file
* Converst the `AAcorrdion` component test to TypeScript

## Notes

First, I noticed that test files are not excluded when compiling TypeScript, which also means that the test files are being shipped to consumers of Magna React. Bundle size concerns aside, this did introduce some difficulty when first migrating the test files to Cypress.

This is because in order to get keywords like `describe`, `it`, and `cy.mount` recognized by TypeScript, [I needed to add the `cypress` directory to the project's root's `include` key inside `tsconfig.json`](https://github.com/cisco-sbg-ui/magna-react/pull/785/commits/62ec3ff357625303ae65b948c9b0af0bd820eb12). After doing that, types in the `**.cy.tsx` files were working properly (and also running properly locally).

However, upon pushing the changes, [Vercel was unable to deploy the Magna React documentation site](https://vercel.com/securex/magna-react/727QD1v5rbRC929zZ5EURX25c7Jd?filter=errors); Vercel logged the following:

```log
[16:32:05.941] > next build
[16:32:05.941] 
... skipping some lines ...
[16:32:13.573] Failed to compile.
[16:32:13.573] 
[16:32:13.573] ./framework/components/AAccordion/AAccordion.cy.tsx:8:1
[16:32:13.573] Type error: Cannot find name 'describe'. Do you need to install type definitions for a test runner? Try `npm i --save-dev @types/jest` or `npm i --save-dev @types/mocha`.
```

I believe this is to do with the fact that when Next.js is building the application for production, it *does not* install the `devDependencies` in `package.json` (as to be expected), and hence, does not have a reference to the TypeScript types provided by Cypress when parsing the test files.

To resolve this issue, I did two things:

(1) Remove `cypress` from the project root's `include` key in `tsconfig.json`
(2) Create another `tsconfig.json` file in the `cypress` directory and put all TypeScript tests under this directory

Now, when Next.js builds the application, it no longer needs to compile a TypeScript file that has a dependency on something listed in `devDependencies` since those files are now underneath the `cypress` directory.

----

At first, I thought adding the test files to the `exclude` key in `tsconfig.json` would resolve the build issue since it would instruct TypeScript to not try compiling the test files to begin with (and thus resolve the Vercel build issue). But then I realized that the consequence of this is that we don't get any TypeScript validation when coding locally. This is what ultimately led me to adding tests in the `cypress` directory.

## References

* [Cypress TypeScript documentation](https://docs.cypress.io/guides/tooling/typescript-support#Clashing-types-with-Jest)
* [Cypress Component Testing FAQ](https://docs.cypress.io/guides/component-testing/faq#Why-isnt-TypeScript-recognizing-the-global-Cypress-objects-or-custom-cypress-commands-eg-cymount)
* [Cypress example to add custom command (e.g. `cy.getByDataTestId`) in TypeScript](https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/fundamentals__add-custom-command-ts)
* [Cypress Jest and TypeScript example ](https://github.com/cypress-io/cypress-and-jest-typescript-example/tree/master)
  * This was useful in determining how to configure the `tsconfig.json` file in the `cypress` directory


cypress/tsconfig.json
brennarvo, line 1: Source: https://github.com/cypress-io/cypress-and-jest-typescript-example/blob/master/cypress/tsconfig.json

androche (approved):
Thanks for doing this @brennarvo!!!

PR #791, opened 2024-07-19
Add auto menu flip logic on bounds checking

rwharrisjr:
Adds a hook, enabled on `ATooltip`, that checks the outer bounds for `AMenu` placement, and flips it it to the opposite if it would render out of bounds.

This only applies to the starting placement - top, right, bottom, left. It does not affect the secondary, in placements such as `top-right`.

### DIFF
--- a/framework/components/AMenuBase/AMenuBase.tsx
+++ b/framework/components/AMenuBase/AMenuBase.tsx
@@ -265,7 +265,6 @@ const AMenuBase = forwardRef<HTMLElement, AMenuBaseProps>(
     const [pointerLeft, setPointerLeft] = useState<number | null>(null);
     const [pointerTop, setPointerTop] = useState<number | null>(null);
 
-    // @ts-ignore
     const {checkMenuSpacing, menuPlacement}: AMenuFlip = useMenuFlip(
       useFlipLogic,
       anchorRef,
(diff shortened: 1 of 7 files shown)


framework/components/AMenuBase/AMenuBase.tsx
```diff
@@ -261,13 +265,31 @@ const AMenuBase = forwardRef<HTMLElement, AMenuBaseProps>(
...
+    const {checkMenuSpacing, menuPlacement}: AMenuFlip = useMenuFlip(
+      useFlipLogic,
+      anchorRef,
+      combinedRef,
+      wrapRef,
+      pointer,
+      placement
+    );
+
+    console.log("placement", placement, menuPlacement);
```
androche, line 278: log

framework/components/AMenuBase/AMenuBase.tsx
```diff
@@ -261,13 +265,31 @@ const AMenuBase = forwardRef<HTMLElement, AMenuBaseProps>(
     const [pointerLeft, setPointerLeft] = useState<number | null>(null);
     const [pointerTop, setPointerTop] = useState<number | null>(null);
 
+    // @ts-ignore
```
androche, line 268: can remove this ts-ignore

androche (commented):
Pulled and tested.  Just two minor things and it seems good to go!

Screenshot 2024-07-22 at 4 48 03 PM

androche (approved):
Looks good, i'd rebase to be safe and then good to merge!

Issue #794, opened 2024-07-19
Issues with AMultiselect


guptar8jan:
**Describe the bug**

1. "required" only works if value prop is set to null. It should work with empty array.
2. "Required" validation do not reset when AMultiselect is disabled. Running `formref.current.reset()` does not help.
3. when modal is scrolling AMultiselect -> dropdown scrolls as well.
<img width="384" alt="image" src="https://github.com/user-attachments/assets/a7e5470c-01ff-4135-9191-806b4cdf7e93">

**To Reproduce**

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - Device: [e.g. Desktop, iPhone6]
 - OS: [e.g. iOS, Ubuntu]
 - Browser [e.g. Chrome, Safari]

**Additional context**

rwharrisjr:
https://github.com/cisco-sbg-ui/magna-react/pull/795 addresses items 1 and 2. I *think* 3 may be addressed by other changes for tracking the menu positioning better, but we may need to add additional scroll checking. This will be easier to test after #792

PR #795, opened 2024-07-19
794 - don't validate disabled fields in forms; validate required empty array for multiselects

rwharrisjr:
Fixes for #794 

Changes

- `AForm` will no longer validate disabled fields. 
- `AForm` will reset the validation state if a field becomes disabled and the form is re-validated
- `AMultiSelect` now validates `required` as non-empty value AND non-empty array. Array is the only valid type for the `value`
- `ASelect` will be properly reset when reset function is called


androche:
@rwharrisjr I've added @guptar8jan as the best person to validate in https://magna-react-difw4mr75.securex-preview.app/- code looks good, but i don't trust that I would test this correctly.

androche (approved):
Reviewed by rekha as well

PR #800, opened 2024-07-23
Set panel id to 0 so it triggers incrementing

rwharrisjr:


androche (approved):

guptar8jan (approved):
why is lib version of AAcoordianPanel different from framework version?

androche:
@guptar8jan lib should have additional files, including d.ts, but the code in AAccordianPanel should be the same.  Not sure what the issue is you're seeing, but maybe an `npm ci` might help - reach out though if you are seeing any other issues!

PR #802, opened 2024-07-23
Remove padding for a-card-container inside modals

rwharrisjr:
Before

Screenshot 2024-07-23 at 2 10 28 PM

AFter

### DIFF
--- a/framework/components/AModal/AModal.js
+++ b/framework/components/AModal/AModal.js
@@ -1,4 +1,11 @@
-import React, {forwardRef, useContext, useEffect, useRef, useMemo} from "react";
+import React, {
+  forwardRef,
+  useContext,
+  useEffect,
+  useRef,
+  useMemo,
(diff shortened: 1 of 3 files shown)


androche (changes requested):
Let's investigate a bit more.  I dont see any issues with this in IM, in fact taking away all padding for a-container in a modal would look like this:
Screenshot 2024-07-23 at 5 01 34 PM

androche:
ah I see the extra padding coming from child elements now.  I _think_ that means it would be safe to remove the container padding because it should just be isolated to modals only.  Although, curious if we can come up with an alternative to editing each elements padding?

androche (approved):
Blocking for the time being so we can get more info.  This might take away padding for other groups so tracking down the problem source a bit more.

androche (approved):
🎉

PR #805, opened 2024-07-24
Reposition open menus on scroll of a container

### LINKED ISSUE #794: Issues with AMultiselect

rwharrisjr:
Resolves #794 

Adds a scroll detection hook, and recalculates menu position when any container is scrolled iff the menu is open.

The handler to recalculate position is throttled to 60fps on window animation frame to reduce the amount of rerenders

### DIFF
--- a/framework/components/AMenuBase/AMenuBase.tsx
+++ b/framework/components/AMenuBase/AMenuBase.tsx
@@ -1,5 +1,6 @@
 import React, {
   forwardRef,
+  useCallback,
   useContext,
   useEffect,
   useRef,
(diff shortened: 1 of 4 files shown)


androche (approved):
NICE!!! Also fixes the minimized window one

PR #807, opened 2024-07-25
Sync hints prop types across componnets; fix medium warning for AInputBase

rwharrisjr:
- Syncs hints to accept string/node everywhere
- Resolves "Warning: Received `true` for a non-boolean attribute `medium`."


androche (commented):
Does this need to be fixed in any of the new typescript type files?   Also we should create a type for this to reuse(future reference)

rwharrisjr:
I don't think it affects any typescript files. In the future a type to share would be good for sure

androche (approved):

PR #808, opened 2024-07-26
Add interactive prop for ACardBasic

rwharrisjr:
Screenshot 2024-07-26 at 10 33 35 AM
Screenshot 2024-07-26 at 10 33 42 AM

### DIFF
--- a/framework/components/ACard/ACard.scss
+++ b/framework/components/ACard/ACard.scss
@@ -173,15 +173,19 @@
 
     &--interactive {
       box-shadow: 0px 2px 5px 0px rgba(0, 0, 0, 0.05);
+      cursor: pointer;
 
       &:hover {
-        border: $border-width solid var(--control-border-hover);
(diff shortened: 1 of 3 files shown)


framework/components/ACard/ACard.scss
```diff
@@ -170,5 +170,25 @@
         background: var(--control-bg-disabled);
       }
     }
+
+    &--interactive {
```
androche, line 174: Lets add the mixin for focus: `@mixin a-control-focus` so we can get that inner state:

androche: Actually differences in mockups -lets keep to stick with figma

androche: Add `cursor: pointer`

androche: Hover color is: border: 2px solid var(--control-border-weak-hover);

androche (approved):
🎉

PR #812, opened 2024-07-28
refactor: types for all

androche:
I know this is wild, and I cant really see straight. BUT this eliminates the need for typescript teams to disable `any` checking in their envs which would have not been ideal.  Instead, we are adding the d.ts files ourselves, fast tracking this project while we move our internal typescript over and convert over to generated d.ts files. 

Please test on as many envs as possible!!!


framework/components/ATriggerTooltip/ATriggerTooltip.tsx
```diff
@@ -30,29 +29,29 @@ const ATriggerTooltip = ({
   const tooltipAnchorRef = anchorRef || triggerRef || firstChildRef;
   const tooltipRef = useRef<HTMLElement>(null);
 
-  const checkForTruncation = useCallback(() => {
-    if (!onlyIfTruncated) {
-      return true;
-    }
+  // const checkForTruncation = useCallback(() => {  //This does not get used in the library
```
androche, line 32: @rwharrisjr  Once the hook was changed to ts, it flagged that `checkForTruncation` is not allowed.  I checked the hook and sure enough it is not there.  If it supposed to be used for `canOpen?` in `useToggle`, these need to match, but i'm also not sure how they would be related.  If its just a way or "retriggering" something - `checkForTruncation` needs to be reassigned to a variable `canOpen`. Can you advise?

rwharrisjr (approved):

PR #818, opened 2024-08-01
Use useEscapeKeyDown for drawer

rwharrisjr:
usePopupQuickExit seems to cause conflicts with existing custom slide in/out controllers for ADrawer. 

Reduce the scope down to just handle escape for accessibility, and let products handle focus accessibility on their own if they aren't using `useDrawerToggle`


androche:
Opened separate ticket here: https://github.com/cisco-sbg-ui/magna-react/issues/819

PR #821, opened 2024-08-01
fix: fix tooltip pointer position

androche:
Screenshot 2024-08-01 at 11 38 14 AM

### DIFF
--- a/framework/utils/helpers.js
+++ b/framework/utils/helpers.js
@@ -1,9 +1,6 @@
 import React from "react";
 
 export const getRoundedBoundedClientRect = (el) => {
-  if (!el.getBoundingClientRect) {
-    return null;
-  }
   const rect = el.getBoundingClientRect();
   return {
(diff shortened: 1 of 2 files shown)


framework/utils/helpers.js
```diff
@@ -1,9 +1,6 @@
 import React from "react";
 
 export const getRoundedBoundedClientRect = (el) => {
-  if (!el.getBoundingClientRect) {
-    return null;
-  }
```
brennarvo, line 6: Would this still be good to keep in here just in case somebody passes a bad `anchorRef`?

androche: Yeaa that was my original thought when adding it, but I took it back out because I got a little nervous  (I tried to change as little component logic as I could over the TS changes).   I think putting it back in should be pretty harmless though - Thanks for bringing it up!

brennarvo (approved):

PR #822, opened 2024-08-02
chore: start of vetting types.

androche:
#771 
This PR examines any default values - while looking through, was able to add some missing params that caught my eye.  Also added paginator type and exporting of all types for other libraries to utilize in their own projects.

### DIFF
--- a/framework/components/AMultiSelect/types.ts
+++ b/framework/components/AMultiSelect/types.ts
@@ -121,7 +121,7 @@ export type AMultiSelectProps = Override<
      */
     filterFunction?: (...args: unknown[]) => unknown;
     /**
-     * Empty state message
+     * Empty state message - NOTE: custom strings should be provided through an i18n library
      *
      * @defaultValue `"No matches found"`
      */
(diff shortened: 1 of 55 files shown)


framework/components/AMultiSelect/types.ts
```diff
@@ -111,11 +112,37 @@ export type AMultiSelectProps = Override<
...
+     *
+     * @defaultValue `[]`
      */
     value?: unknown[];
     /**
      * Function to filter items when the input value changes
      */
     filterFunction?: (...args: unknown[]) => unknown;
+    /**
+     * Empty state message
```
rwharrisjr, line 124: Could we add a "NOTE: ..." here to say "custom strings should be provided through an i18n library"?

framework/components/APaginator/types.ts
```diff
@@ -1 +1,59 @@
...
+     */
+    onPageChange?: (newPage: number) => void;
+    /**
+     * Handles the submission of results per page values.
+     */
+    onResultsPerPageChange?: (resultsPerPage: number) => void;
+    /**
+     * Sets the current page.
+     *
+     * @defaultValue `0`
```
rwharrisjr, line 17: Could we add an extra note to make it clear that APaginator indexes at 0?

rwharrisjr (changes requested):

PR #824, opened 2024-08-05
Add boolean defaults, more docs

rwharrisjr:
Does a pass at more defaults/types/docs through `ADrawer`


androche (approved):

androche:
Added to issue tracking: #771

PR #830, opened 2024-08-08
Feat: Allow custom placeholder components

jonathanmcsweet:
- [x] The changes are documented in component docs

Feature: Updated ASelect placeholder prop to take a component or a string and also added a code example

- Currently the placeholder prop only accepts a string although you can pass a component but will get an error in the browser console

- The new behavior allows for the placeholder prop on ASelect to render a custom component or a string

- This should be entirely backwards compatible with previous use cases


rwharrisjr:
@jonathanmcsweet  the type in `types.ts` needs to be updated as well

framework/components/ASelect/ASelect.js
```diff
@@ -341,6 +341,13 @@ const ASelect = forwardRef(
 
     const selectIcon = isOpen ? "chevron-up" : "chevron-down";
 
+    const eitherPlaceholder =
```
rwharrisjr, line 344: small nitpick on magna code patterns - other components will do something like

`placeholder: propsPlaceholder` in the component props, then `const eitherPlaceholder` could be `const placeholder`

rwharrisjr (changes requested):

framework/components/ASelect/ASelect.js
jonathanmcsweet: - [ ] Use standard convention for props placeholder

rwharrisjr (approved):

PR #831, opened 2024-08-08
Remove logic around scrollLeft/scrollTop for menu placement

rwharrisjr:
Previously the code used `appCoords.scrollLeft` and `appCoords.scrollTop` to calculate offsets of the wrapper when not using `withNewWrappingContext`. 

However, both of those values would always be undefined, as they aren't part of `DOMRect`.

This caused every `AMount` container without `offsetParent` of `document.body` to use undefined values in the offset calculation. This led to the values always computing to `NaN`.

The values being used were changed in https://github.com/cisco-sbg-ui/magna-react/pull/773/files#diff-bcc8104da7921ce69a5415f3862dc9c690e7515fee36da98867d8bec28feae41R148

By changing `appCoords.scrollLeft` and `appCoords.scrollTop` to `appCoords.left` and `appCoords.top`, the calculations started working, but were using values that were often positioned relative to the page vs relative to the intended position of the menu around the anchor.

---------

This PR removes the code that would never actually work, and conditionalized the page level logic to retain its functionality.

### DIFF
--- a/framework/components/AMenuBase/AMenuBase.tsx
+++ b/framework/components/AMenuBase/AMenuBase.tsx
@@ -38,7 +38,6 @@ const calculateMenuPosition = ({
     return;
   }
 
-  const appCoords = getRoundedBoundedClientRect(appRef.current);
   const wrapCoords = getRoundedBoundedClientRect(wrapRef.current);
   const anchorCoords =
     anchorRef instanceof DOMRect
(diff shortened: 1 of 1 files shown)


androche:
🎉   Really good catch, also kudos to typescript paying some dividends there. 😄 

If you could post steps on where I could verify that would be helpful. I'm unable to recreate locally and havent ran into it in incident manager as I click around.

framework/components/AMenuBase/AMenuBase.tsx
```diff
@@ -147,38 +147,38 @@ const calculateMenuPosition = ({
     pageHeight = wrapCoords.height;
   } else {
     // Calculate positioning relative to window or app/wrap elements
-    xOffset = appRefEl.offsetParent?.isSameNode(document.body)
-      ? window.pageXOffset
-      : wrapCoords.left - appCoords.left; // this was incorrectly set to "scrollLeft" when scrollLeft is not a return value in appCoords
+    if (appRefEl.offsetParent?.isSameNode(document.body)) {
+      xOffset = window.pageXOffset;
```
androche, line 151: Can delete `appCoords` now

Issue #789, opened 2024-07-18
ALoader issue - variant progressbar


guptar8jan:
**Describe the bug**

ALoader issue - variant progressbar: only works with label.
WORKS:
<ALoader variant="progressbar" size="medium" indicatorProps={{indeterminate: true}}>Medium progress bar</ALoader><br />
 NO SHOW:
<ALoader variant="progressbar" size="medium" indicatorProps={{indeterminate: true}}></ALoader><br />
      
**To Reproduce**

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - Device: [e.g. Desktop, iPhone6]
 - OS: [e.g. iOS, Ubuntu]
 - Browser [e.g. Chrome, Safari]

**Additional context**

androche:
I went to create the new component and it said it was already made!
https://magna-react.vercel.app/components/progressbar
Looks like we have another component for this.  @guptar8jan could you try with this one in the meantime to see if it fixes the issues and we will continue work adding the variants and such.

androche:
Added in: https://github.com/cisco-sbg-ui/magna-react/pull/839

PR #839, opened 2024-08-16
Progressbar updates/#789

androche:
**#789** Refactor and redesign of the progressbar component to match magnetic.

_(breaking changes)_

**This is up for review BUT we may need to wait a version for it.  It should go in with other breaking changes**

Update 8/20 - after a progress bar audit, it appears the only breaking change for the current usage in other apps would be a group using `striped` prop which has a name change of `animate`.

Screenshot 2024-08-16 at 11 38 39 AM

Screenshot 2024-08-16 at 11 38 47 AM


rwharrisjr (approved):
colorful

PR #848, opened 2024-08-22
Add accessibility to interactive tags

rwharrisjr:
Tags should be accessible if they have an `onClick` set.

When interactive, the tag will now get the element `<button>`, and pick up a focus state

Screenshot 2024-08-22 at 4 59 28 PM


androche (changes requested):
Nice!  I dont think ppl know we accept an `onClick` though (I had to look it up myself 😄 ) as its not in docs or props.  We should add an example and update types as well as prop-types!

androche (approved):

androche (changes requested):
Oops, approved just reading one commit on the ts config- putting back to review so I can take a good look.

rwharrisjr:
Replaced by #854

androche:
and https://github.com/cisco-sbg-ui/magna-react/pull/851

PR #851, opened 2024-08-28
A tag test

androche:
Reads props, but not sure if it has any side effect with usage - needs testing in typescript environments.


rwharrisjr (approved):

androche:
Tested across 3 typescript repos and this actually also seemed to solve the type checking issue as well with polymorphic components!

mvelkCisco:
@androche careful here, this does not solve the polymorphism issue, it breaks it. Props of the components passed to the `component` prop are no longer respected. 

I think it's because you are basically statically setting the `C` argument. That needs to stay empty so TS can infer props itself based on the `component` prop.

androche:
> @androche careful here, this does not solve the polymorphism issue, it breaks it. Props of the components passed to the `component` prop are no longer respected.
> 
> I think it's because you are basically statically setting the `C` argument. That needs to stay empty so TS can infer props itself based on the `component` prop.
> 
> image

Thanks for this catch @mvelkCisco!  I knew it seemed too easy! 😄

PR #854, opened 2024-08-28
Tag focus/hover fixes; hideStatusIcon prop

rwharrisjr:
- Adds the standard button control focus/hover states to "interactive" tags
- Adds `hideStatusIcon` prop so tags like info/disabled can be used in more contexts 
Screenshot 2024-08-28 at 2 49 24 PM
- Fixes hover border-radius on interactive tags

@wlorand3 fyi

### DIFF
--- a/framework/components/ATag/types.ts
+++ b/framework/components/ATag/types.ts
@@ -48,6 +48,8 @@ export type ATagProps<C extends React.ElementType> =
       customIcon?: boolean | React.ReactNode;
       /**
        * Hide the status icon
+       *
+       * @defaultValue `false`
        */
       hideStatusIcon?: boolean;
       "data-testid"?: string;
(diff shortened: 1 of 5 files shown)


framework/components/ATag/types.ts
```diff
@@ -46,6 +46,10 @@ export type ATagProps<C extends React.ElementType> =
        * @defaultValue `false`
        */
       customIcon?: boolean | React.ReactNode;
+      /**
+       * Hide the status icon
+       */
```
androche, line 51: Add `defaultValue` and mdx example if possible

androche:
Is it possible to add the border to just the tag so it doesn't "expand"?  
Screenshot 2024-08-28 at 4 50 06 PM
It looks like it doesn't need a double unless its an "input" style which we dont have here yet I dont thinkg

rwharrisjr:
I reused the control focus mixin so it would be closer to the AButton focus style

androche:
Yea, I think the expanding on tertiary and things like this looks like a bug or something personally. lol   I double checked CDS and they do just the single border on focus if we wanted an excuse.  We could also stick with this method though.

androche (approved):

PR #855, opened 2024-08-28
fix: add info and adjust mdx to latest figma

androche:
Not including `info` in prop list was causing issues in TS repos.  Fixed up mdx and added notes on variant deprecations, however this is still backwards compatible.

Screenshot 2024-08-28 at 4 25 35 PM

### DIFF
--- a/framework/components/AEmptyState/types.ts
+++ b/framework/components/AEmptyState/types.ts
@@ -1,12 +1,24 @@
 import {Override} from "../../types";
 
-export type AEmptyStateVariant = "success" | "positive" | "warning" | "danger";
+export type AEmptyStateVariant =
+  | "success"
+  | "positive"
+  | "warning"
+  | "danger"
+  | "negative"
+  | "info";
 
 export type AEmptyStateProps = Override<
   React.ComponentPropsWithRef<"div">,
   {
     /**
      * Empty state variant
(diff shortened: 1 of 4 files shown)


framework/components/AEmptyState/types.ts
```diff
@@ -1,12 +1,24 @@
...
+  | "info";
 
 export type AEmptyStateProps = Override<
   React.ComponentPropsWithRef<"div">,
   {
     /**
      * Empty state variant
+     * @defaultValue `"info"`
+     *
+     * deprecations - use "negative" instead of "danger"
```
rwharrisjr, line 18: would it be better to use `@deprecated` so it'll highlight?

rwharrisjr (changes requested):

framework/components/AEmptyState/types.ts
androche: Well the whole prop isn't deprecated and it would show everyone a line across "variant".  Unfortunately, you cant add deprecations to individual union types.

rwharrisjr (approved):

androche:
@rwharrisjr I was able to adjust line-height in message, but still had to add some margin to space it out.

rwharrisjr (approved):

PR #857, opened 2024-08-29
chore: add back .js file

androche:
@rwharrisjr re: @mvelkCisco response in https://github.com/cisco-sbg-ui/magna-react/pull/851#issuecomment-2316997714

Adding back the .js file now for rendering and docs generation while we work on this more.  Tsx can stay so types get picked up at least.  If there were any changes added in the `ATag.tsx` they will also need to get added to this `ATag.js` as it is the previous version of what went out in and due to file change was unable to get a diff:
https://github.com/cisco-sbg-ui/magna-react/pull/848
https://github.com/cisco-sbg-ui/magna-react/pull/851

### DIFF
--- a/framework/components/ATag/ATag.js
+++ b/framework/components/ATag/ATag.js
@@ -1,10 +1,13 @@
+import PropTypes from "prop-types";
 import React, {forwardRef} from "react";
 import AIcon from "../AIcon";
 
 import {keyCodes} from "../../utils/helpers";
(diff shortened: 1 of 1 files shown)


rwharrisjr (approved):

androche:
Will rebase in https://github.com/cisco-sbg-ui/magna-react/pull/854, and apply changes to .js file over there

PR #859, opened 2024-08-30
Floating UI POC - Dropdown menus

rwharrisjr:
This is a POC to explore the feasibility of replacing the positioning logic in `AMenuBase` (which is error prone) with a well designed and maintained positioning library. 

https://floating-ui.com/

Floating UI was designed to handle almost every situation of positioning elements on top of another. It is the successor to TippyJS and PopperJS.

------------

This would break the longstanding precedence of no dependencies in this project, but the benefits are possibly worth it. 

- Allows for controlling offsets easily
- Easy to tie in react portals to handle situations like Modals
- Positions should be less flakey
- Reduces overall code
- Is flexible enough to replace positioning logic without replacing component logic

------------

This PR reworks 

- `AMultiSelect`
- `ASelect`
- `AAutocomplete`
- `ACombobox`

These changes are non-breaking, and does not affect other components that use `AMenu/AMenuBase`. This also improves some of the keyboard event handling and menu focus handling.


androche (approved):
I've pulled and tested this in monorepo and confirmed it fixes scroll with page!!  Flipping also works nicely, but flipping also appears to be working as we currently have it -i'm sure it fixes a case I couldnt recreate.

PR #865, opened 2024-09-10
Let next have its own tsconfig file

rwharrisjr:
This should keep tsconfig.json from getting overwritten when we start the dev server


androche (approved):
YAYYYYYY!!!!!

PR #872, opened 2024-09-12
Stop checking tab overflow once overflow is hit

rwharrisjr:
If tab n+1 is smaller than tab n, and tab n overflows but tab n+1 doesn't, n+1 can show and cause the list to be out of order


androche (approved):
ATab lives to see another day!

PR #874, opened 2024-09-16
fix: fixes new random css ordering issue in amultiselect

androche:
Note* this is an issue only after deployment

**Before **
Screenshot 2024-09-16 at 11 52 10 AM

**After **
Screenshot 2024-09-16 at 11 53 15 AM

@radesai21

### DIFF
--- a/framework/components/AFieldBase/AFieldBase.scss
+++ b/framework/components/AFieldBase/AFieldBase.scss
@@ -2,8 +2,6 @@
 
 .a-field-base {
   position: relative;
-  display: inline-flex;
-  flex-direction: column;
   width: 100%;
 
   &__label {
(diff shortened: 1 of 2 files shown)


androche:
@rwharrisjr instead of important we could also do `unset: width` here: https://github.com/cisco-sbg-ui/magna-react/blob/main/framework/components/AMultiSelect/AMultiSelect.scss#L4


But i'm honestly not sure what impacts that would have in certain cases?  Seems like a certain risk either way.

androche:
Re-verified that this is still the reason we added `width: unset;`

https://github.com/cisco-sbg-ui/magna-react/pull/419/files#diff-8450920dd7c0985ea648799f082ade3b5794f88c85d10d2313b8d3059d520f65R35

rwharrisjr (approved):

rwharrisjr (approved):

PR #896, opened 2024-09-20
ACopyButton - Don't apply button icon style when there's text

rwharrisjr:
Screenshot 2024-09-20 at 11 33 20 AM

https://github.com/cisco-sbg-ui/magna-react/commit/d17d584bc8a669eeedec41d7610f222f59489770#diff-966cc622ccfa356f0b786b6d2e927430f02cf877c24b3878c30b5879a152cb9cR149

Not sure what else that may break

### DIFF
--- a/framework/components/ACopyButton/ACopyButton.js
+++ b/framework/components/ACopyButton/ACopyButton.js
@@ -55,7 +55,7 @@ const ACopyButton = forwardRef(
         <AButton
           ref={combinedRef}
           className={className}
-          icon
+          icon={!children && !defaultLabel}
           onClick={() => {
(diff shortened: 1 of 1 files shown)


androche (approved):
In `AButton` the `icon` prop is and has only been for icon only buttons.   It was actually never meant to serve both icons with text and single icons, but "worked" because we never applied a width.  There's styling differences between icon-only and icon-with-text buttons including the sizes of the icons themselves, so once we set the appropriate widths to icon-only buttons, it now accurately shows the difference between the two.   The issues i've seen around is that devs will need to remove "icon" from buttons with text as it's a misuse of the prop.

PR #903, opened 2024-09-30
Input disabled state should take precendence over readOnly

rwharrisjr:
```
<AInputBase
  append="[Append]"
  prepend="[Prepend]"
  className="mb-3"
  clearable
  readOnly
  disabled
  onClear={() => alert("cleared")}
  label="Label - Focused"
  hints="This is a sample of hint text below a form input"
  focused
>
  [Children]
</AInputBase>
```

Try on Extend -> Input Base. Note the input's background color. Disabled should gray it out even if `readOnly` is set


androche (approved):
Pulled and tested

PR #914, opened 2024-10-08
Convert APopover to use AFloatingBase

rwharrisjr:


androche (changes requested):
Good news is it works 😄 
A few issues came up in other envs:

1. Rebase and run `npm run lint` not sure what will show up there.  This will also give the labels fixes I believe for oak that keep failing. 
2. Run in oak from there and there appears to be a role menu issue where the element is now nested - wonder if we can bring it to the top level or if the tests need to be changed
Failing on getbyrole menu
Screenshot 2024-10-11 at 9 26 31 AM

Current implementation on INT
Screenshot 2024-10-11 at 9 42 27 AM

3. IM has some failures around assigneeAvatarTooltip which I thought we looked into last week, but this does appear to be a new error. 

4. We might consider taking out the console.warn for toast as the warning is triggering test errors.  Maybe a console.log in that case would still be good.

5. Tooltip vs popover issues:

Screenshot 2024-10-11 at 9 18 29 AM
Screenshot 2024-10-11 at 9 17 57 AM

rwharrisjr:
`npm run lint` doesn't return any issues

framework/components/APopover/APopover.d.ts
```diff
@@ -3,7 +3,7 @@ import React from "react";
 import {type APopoverProps} from "./types";
 
 declare const APopover: <C extends React.ElementType = "div">(
```
androche, line 5: Getting a C is defined but never used

androche (approved):
Approving ahead of time, but see notes below before merging! 
- Remove C type
- IM is passing!  🎉 
- Oak still has same menu error so not sure what's going on there?  Might be worth merging and seeing what's left over there after the toast fix below:
- If you could also sneak in the console.log on the toast here that would be good too.

androche (approved):

PR #917, opened 2024-10-09
Add AEmptyState horizontal layout; align magnetic small/medium/large sizes

rwharrisjr:
Magnetic figma shows smaller icon sizes and a larger gap under the label. Note this does not align with current CDS implementation, but is less comically large than before. 

Screenshot 2024-10-09 at 5 03 51 PM

### DIFF
--- a/framework/components/AEmptyState/AEmptyState.js
+++ b/framework/components/AEmptyState/AEmptyState.js
@@ -5,6 +5,12 @@ import React, {forwardRef} from "react";
 import AIcon from "../AIcon";
 import "./AEmptyState.scss";
 
+const baseClass = "a-empty-state",
+  backgroundClass = `${baseClass}__background`,
+  labelClass = `${baseClass}__label`,
+  messageClass = `${baseClass}__message`,
+  messageContainerClass = `${baseClass}__message-container`;
+
 const AEmptyState = forwardRef(
   (
     {
@@ -20,14 +26,11 @@ const AEmptyState = forwardRef(
       medium,
       large,
       xlarge,
+      horizontal = false,
       ...rest
     },
     ref
   ) => {
-    const baseClass = "a-empty-state",
-      backgroundClass = `${baseClass}__background`,
-      labelClass = `${baseClass}__label`,
-      messageClass = `${baseClass}__message`;
     let className = baseClass,
       iconClass = `${baseClass}__icon`,
       containerClass = `${baseClass}__container`,
@@ -55,6 +58,10 @@ const AEmptyState = forwardRef(
       icon = "info";
     }
 
+    if (horizontal) {
+      className += ` ${baseClass}--layout-horizontal`;
+    }
+
     if (xsmall) {
       className += ` ${baseClass}--xsmall`;
     } else if (small) {
@@ -75,9 +82,11 @@ const AEmptyState = forwardRef(
           <AIcon className={backgroundClass}>empty-background</AIcon>
           <AIcon className={iconClass}>{propsIcon || icon}</AIcon>
         </div>
-        {label && <div className={labelClass}>{label}</div>}
-        {message && <div className={messageClass}>{message}</div>}
-        {children}
+        <div className={messageContainerClass}>
+          {label && <div className={labelClass}>{label}</div>}
+          {message && <div className={messageClass}>{message}</div>}
+          {children}
+        </div>
       </div>
     );
   }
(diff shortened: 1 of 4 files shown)


framework/components/AEmptyState/AEmptyState.js
```diff
@@ -5,6 +5,12 @@ import React, {forwardRef} from "react";
 import AIcon from "../AIcon";
 import "./AEmptyState.scss";
 
+const baseClass = "a-empty-state",
+  backgroundClass = `${baseClass}__background`,
+  labelClass = `${baseClass}__label`,
+  messageClass = `${baseClass}__message`,
+  messageContainerClass = `${baseClass}__message-container`;
+
```
brennarvo, line 13: Nice refactor moving this out of the component 👍

framework/components/AEmptyState/AEmptyState.scss
```diff
@@ -33,9 +35,38 @@ $container-xl: 400px;
...
   }
 
+  &__message-container {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+  }
+
+  &__label {
+    margin: 0 0 12px 0;
```
brennarvo, line 45: Is the bottom-margin here coincidental with the `$icon-xs` sass variable? Only wondering in case it should always be synced with the size of the small icon.

brennarvo (approved):
This looks good to me. @androche, you might want to double check me on the icon resizing though

framework/components/AEmptyState/AEmptyState.scss
rwharrisjr: coincidental, that just seemed to be the gap that the figma had. Combined with the line-height for each size it fit

androche (changes requested):
Oops - sorry i missed this message last week.

Ok so to be specific:

Technically, starting with a medium size, the icon size should be (according to figma) 52.93.  BUT the catch is our icon is actually smaller than the figma even though we might specify 53px in css is clocking in  at `43px`.
Screenshot 2024-10-14 at 10 34 36 AM

We'll need to adjust our css to get it as close to 53 as possible, which in our code is `$icon-md: 65px;` and the surrounding icons and sizes need to be adjusted to this.
Screenshot 2024-10-14 at 10 37 45 AM


@rwharrisjr happy to help gather the rest of these numbers while you make these adjustments to medium


Update - I was viewing the icon in the sidebar which still included additional padding so these number are incorrect so please disregard - I sent over adjustments with correct pxs.

framework/components/AEmptyState/AEmptyState.scss
```diff
@@ -172,11 +207,13 @@ $container-xl: 400px;
     .a-empty-state__label {
       font-weight: 700;
       font-size: 24px;
```
androche, line 209: `font-size: 18px`

framework/components/AEmptyState/AEmptyState.scss
```diff
@@ -172,11 +207,13 @@ $container-xl: 400px;
...
+      line-height: $line-height--xl;
     }
 
     .a-empty-state__message {
       font-weight: 400;
       font-size: 18px;
```
androche, line 215: `font-size: 16px"

framework/components/AEmptyState/AEmptyState.scss
```diff
@@ -33,9 +35,38 @@ $container-xl: 400px;
...
+    flex-direction: row;
+
+    .a-empty-state__container {
+      flex: 0 0 auto;
+    }
+
+    .a-empty-state__message-container {
+      align-items: flex-start;
+
+      margin-left: 12px;
```
androche, line 64: this is actually `40px` for large, but this would need to be adjusted for each medium and small

androche: What might be easier is doing margin right adjustments on the icon container itself since we already have code separated out for that. 

.a-empty-state--large .a-empty-state__container {
    height: 224px;
    width: 224px;
    margin-right: 40px;
}

framework/components/AEmptyState/AEmptyState.scss
androche: On horizontal this is actually 8px on the large going smaller with each size.    It might be better to do label space in the sizing css already set up (mentioned below) and just override this if you didn't want to spend more time on it.

androche (changes requested):

framework/components/AEmptyState/AEmptyState.scss
androche: Decreasing the margin for each size variation

androche (approved):
Looks great!!

@brennarvo please ignore extra review prompt - I clicked the wrong button! :)

PR #918, opened 2024-10-14
Convert ADropdown to use AFloatingMenu

rwharrisjr:
Additionally adds ability to set `component` of either `ATag` or `AButton`
Extends `ATag` to handle a `dropwdown` variant


framework/components/ATag/ATag.js
androche, line 1: We kept this file in for props.  Otherwise they are empty
Screenshot 2024-10-15 at 9 09 07 AM

rwharrisjr: Ah, I thought we might have fixed that

androche:
Cool although with this approach can the tags still be different colors and variations?  I'm wondering if it makes more sense to fix up the tag itself?
Screenshot 2024-10-15 at 2 23 45 PM

rwharrisjr:
@androche maybe if we added in a `componentProps` prop? Either that or we can split it out into ATagDropdown, but that seems excessive

PR #927, opened 2024-10-16
Refactor to place FloatingFocusManager below FloatingPortal

rwharrisjr:


framework/components/AFloatingMenu/AFloatingMenu.tsx
```diff
@@ -139,21 +139,22 @@ const AFloatingMenu = forwardRef<
...
     return (
-      <FloatingFocusManager
+      <AFloatingMenuContainer
+        ignoreOutsideClick
         context={context}
-        disabled={!focusOnOpen}
+        focusOnOpen={focusOnOpen}
         initialFocus={initialFocus}>
         <AList
+          ref={ref as any}
```
androche, line 152: Anything we can do here other than any?

framework/components/AFloatingMenu/types.ts
```diff
@@ -1,31 +1,38 @@
...
+   */
+  focusOnOpen?: boolean;
+  initialFocus?: React.RefObject<HTMLDivElement>;
+};
 
 export type AFloatingMenuProps<C extends React.ElementType> = Override<
   AListProps<C>,
   {
-    context?: React.Context<any>;
+    context: FloatingContext<ReferenceType>;
```
androche, line 35: This is now required - just making sure thats correct?

rwharrisjr: Yes. `context` is required for the focus manager to work, and it is always set from menu components since they have to use the `useFloatingDropdown` hook

framework/components/AFloatingMenu/AFloatingMenu.tsx
androche: Well two options could be: 

One we could take out the div in AListProps - so it would be like:

```

declare const AList: <C extends React.ElementType>(
  props: AListProps<C>
) => JSX.Element;

export default AList;
```

Or potentially a safer in the moment option (but not always great long term) would be to cast it as  
`<AList
       ref={ref as React.RefObject<HTMLDivElement>}`

androche (approved):
Tested in multiple envs

PR #929, opened 2024-10-17
Fix FloatingArrow NaN console warning

rwharrisjr:
On occasion it seems the initial placement calculation returns NaN for the X coordinate, and the arrow renders fast enough it tries to use it. 

Adding a check to make sure X is set before rendering the arrow

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@cisco-sbg-ui/magna-react",
-  "version": "2.6.4",
+  "version": "2.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cisco-sbg-ui/magna-react",
-  "version": "2.0.0",
+  "version": "2.6.4",
```
androche, line 3: Undo

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@cisco-sbg-ui/magna-react",
-  "version": "2.0.0",
+  "version": "2.6.4",
```
androche, line 3: `npm i`

androche (approved):

PR #937, opened 2024-10-21
fix flex spacing properties

Graham187K:
This fixes an issue spotted in an Incident Manager PR: https://github.com/advthreat/incident-manager/pull/3009#issuecomment-2427565813

When the `AInput` is narrower (< 220px), it eventually makes the "close" button shrink.

Before fix:
image

After fix:
image


androche (approved):
🎉

rwharrisjr (approved):

PR #938, opened 2024-10-22
Proposal: Responsive Width Drawer

Graham187K:
Figma for reference: https://www.figma.com/design/zYfa9Hqw0D8b5hpuVwqMeQ/Reusable-Drawer-components-FY24?node-id=1823-23370&node-type=rounded_rectangle&m=dev

Magnetic Drawer Width guidelines

Incident manager needs to implement most `ADrawer`s to have a variable width depending on the viewport width. This follows Magnetic guidelines for Drawer width and responsive breakpoints.

|Breakpoint|Viewport|Drawer|
|---|---|---|
|SM|960px|400px|
|md|1280px|400px|
|LG|1680px|576px|
XL|2080px|768px|

This PR implements a new prop `responsiveWidth` on `ADrawer`.

```ts
<ADrawer 
  responsiveWidth={true} 
  {...otherProps}
/>
```

When `responsiveWidth={true}` prop is set, the Drawer width will automatically be responsive. 
For backward compatibility, when the `openWidth` prop is set, then it will supersede the `responsiveWidth` prop and the Drawer will have a fixed width. If neither prop are provided, then the Drawer will remain at it's default width at all times.

### DIFF
--- a/framework/components/ADrawer/ADrawer.scss
+++ b/framework/components/ADrawer/ADrawer.scss
@@ -148,7 +148,7 @@ $common-transitions: transform $transition-props, width $transition-props,
     top: unset;
   }
 
-  &--auto-width {
+  &--responsive-width {
     width: 400px;
 
     @media screen and (min-width: 1680px) {
--- a/framework/components/ADrawer/ADrawer.scss
+++ b/framework/components/ADrawer/ADrawer.scss
@@ -147,6 +147,18 @@ $common-transitions: transform $transition-props, width $transition-props,
     left: 0;
     top: unset;
   }
+
+  &--auto-width {
+    width: 400px;
+
+    @media screen and (min-width: 1680px) {
+      width: 576px;
+    }
+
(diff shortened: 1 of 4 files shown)


framework/components/ADrawer/ADrawer.scss
```diff
@@ -147,6 +147,18 @@ $common-transitions: transform $transition-props, width $transition-props,
     left: 0;
     top: unset;
   }
+
+  &--responsive-width {
+    width: 400px;
+
+    @media screen and (min-width: 1680px) {
```
rwharrisjr, line 154: I think we could do a variable lookup here to keep things consistent if Magnetic changes the breakpoint thresholds. 

`map-get($grid-breakpoints, "lg")`

https://github.com/cisco-sbg-ui/magna-react/blob/main/framework/styles/utilities/variables.scss#L97-L102

rwharrisjr (changes requested):

rwharrisjr (approved):

PR #947, opened 2024-10-25
Add a few OS icons

rwharrisjr:
Screenshot 2024-10-25 at 9 22 23 AM


framework/components/AIcon/iconMap.ts
```diff
@@ -43,13 +44,16 @@ export const iconNameMap: IconMap = {
   comments: "chat-text",
   compare: "square-half",
   computer: "desktop",
+  "computer-tower": "",
```
androche, line 47: hmm I thought we decided we didn't need these?  I haven't been adding anything additional here

rwharrisjr: It was a request for convenience since ones like the apple were being used already

androche: i'm saying adding it to the iconMap - that was for atomic initially and a while back, we deleted anything that didn't map specifically.  I dont think adding them in this map is necessary is all

rwharrisjr: Ah, I just did it so it keeps in sync with the copy script. I doubt the icons will change significantly enough to matter, but it does account for that if so

androche: Ok, just because all 20-30 icons or so i just added didn't go in here lol

androche: Here's where we started deleting - pretty much anything with empty strings it seems.  https://github.com/cisco-sbg-ui/magna-react/commit/c1335696233fe4389c2415d1d83ba027f0723bc6#diff-32b1c8ba1b94aee6a8f0dddec32387ea75a9bf8a81d898daf91938fbe472ee2f

rwharrisjr: Yeah.. I'm kinda going back and forth how to manage it going forward. If it weren't for the styling requirements I'd say we could include phosphor-react as a dep and have a wrapper... but that is impractical

androche (approved):

PR #961, opened 2024-11-13
Allow ADropdown to be controlled externally

rwharrisjr:
Needed for integration with some UI elements that auto open dropdowns

### DIFF
--- a/framework/components/ADropdown/ADropdown.mdx
+++ b/framework/components/ADropdown/ADropdown.mdx
@@ -434,7 +434,7 @@ For more advanced use cases of dropdown see `AMenu` and `AButton`
 
 Combining `isOpen` with `onClose` allows for external control of the open state. The menu will still be interactable with clicks.
 
-Note: any extra component props are passed to the anchor element.
+The menu will respect `true` for `isOpen`, but otherwise handle clicks of menu items and outside click as normal.
 
 <Playground
   code={`() => {
--- a/framework/components/ADropdown/ADropdown.mdx
+++ b/framework/components/ADropdown/ADropdown.mdx
@@ -454,7 +454,7 @@ return (
   <br />
   <ADropdown
     ref={ref}
-    open={isOpen}
+    isOpen={isOpen}
     onClose={() => {
       setIsOpen(false);
     }}
--- a/framework/components/ADropdown/ADropdown.mdx
+++ b/framework/components/ADropdown/ADropdown.mdx
@@ -448,7 +448,7 @@ return (
     onClick={() => {
       setIsOpen(!isOpen);
     }}>
-    Opens the dropwdown
+    Opens the dropdown
   </AButton>
   <br />
   <br />
--- a/framework/components/ADropdown/ADropdown.mdx
+++ b/framework/components/ADropdown/ADropdown.mdx
@@ -430,6 +430,53 @@ For more advanced use cases of dropdown see `AMenu` and `AButton`
 `}
 />
 
+### External open state
+
+Component props are passed to the anchor element
+
+<Playground
+  code={`() => {
+  const ref = useRef(null);
+  const [isOpen, setIsOpen] = useState(false);
+
+return (
+
+<div>
+  <AButton
+    onClick={() => {
+      setIsOpen(!isOpen);
+    }}>
+    Opens the dropwdown
+  </AButton>
+  <br />
+  <br />
+  <ADropdown
+    ref={ref}
+    open={isOpen}
+    onClose={() => {
+      setIsOpen(false);
+    }}
+    title="Has a ref">
+    <AListItem
+      onClick={() => {
+        console.log(ref.current);
+      }}>
+      List item
+    </AListItem>
+    <AListItem
+      onClick={() => {
+        console.log(ref.current);
+      }}>
+      List item
+    </AListItem>
+  </ADropdown>
+</div>
+
+);
+}
+`}
+/>
+
 ### Component props
 
 Component props are passed to the anchor element
--- a/framework/components/ADropdown/ADropdown.mdx
+++ b/framework/components/ADropdown/ADropdown.mdx
@@ -456,7 +456,7 @@ return (
     onClose={() => {
       setIsOpen(false);
     }}
-    title="Has a ref">
+    title="Can be opened by a state prop">
     <AListItem
       onClick={() => {
         console.log(ref.current);
--- a/framework/components/ADropdown/ADropdown.mdx
+++ b/framework/components/ADropdown/ADropdown.mdx
@@ -432,7 +432,9 @@ For more advanced use cases of dropdown see `AMenu` and `AButton`
 
 ### External open state
 
-Component props are passed to the anchor element
+Combining `isOpen` with `onClose` allows for external control of the open state. The menu will still be interactable with clicks.
+
+Note: any extra component props are passed to the anchor element.
 
 <Playground
   code={`() => {
--- a/framework/components/ADropdown/ADropdown.tsx
+++ b/framework/components/ADropdown/ADropdown.tsx
@@ -13,7 +13,7 @@ const ADropdown = forwardRef<HTMLElement, ADropdownProps<React.ElementType>>(
       className: propsClassName,
       component = AButton,
       disabled = false,
-      open,
+      isOpen: open,
       title,
       placement,
       menuClass,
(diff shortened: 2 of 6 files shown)


framework/components/ADropdown/ADropdown.tsx
```diff
@@ -29,14 +31,24 @@ const ADropdown = forwardRef<HTMLElement, ADropdownProps<React.ElementType>>(
     const [isOpen, setIsOpen] = useState(false);
     const Component = component || AButton;
 
+    const openState = open || isOpen;
```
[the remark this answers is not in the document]
rwharrisjr: This logic would change my intention. My idea was that the open state can be controlled by an external prop, but otherwise the menu should respond to the normal close events (click outside, click on button, escape key). 

I think normal close behavior is desired vs having to have the consuming side manage closing of the menu, particularly in instances that some kind of UI automation opens the menu for the user and intends for the user to then interact with the menu

framework/components/ADropdown/ADropdown.mdx
```diff
@@ -430,6 +430,55 @@ For more advanced use cases of dropdown see `AMenu` and `AButton`
 `}
 />
 
+### External open state
+
+Combining `isOpen` with `onClose` allows for external control of the open state. The menu will still be interactable with clicks.
+
+The menu will respect `true` for `isOpen`, but otherwise handle clicks of menu items and outside click as normal.
+
+<Playground
```
androche, line 439: When i'm clicking open one, it doesn't toggle like the others. Could be missing the use case for open only, but seems like it should probably just toggle like the others' behavior:

toggle

androche:
Lastly, if it could just be tested in IM or somewhere that is already using ADropdown and run TS in there that would be good

framework/components/ADropdown/ADropdown.mdx
rwharrisjr: It won't be a button that'll set the open state for the main use case. And in the case of an external button, it would be the consumer's responsibility to toggle the state to false

androche (approved):
approved

mvelkCisco:
I just came across this randomly:

> The menu will respect `true` for `isOpen`, but otherwise handle clicks of menu items and outside click as normal.

Does that mean that `isOpen={false}` will not close the dropdown? Because if so, that's very confusing. 

All we need is the dropdown to respect the `isOpen` fully when it's controlled from outside. Otherwise we will have inconsistent internal state with whatever value is in the `isOpen={myControledBooelan}` and the consumers would need to do some awkward state syncing too.

rwharrisjr:
Setting `isOpen={false}` will close the menu, if it was opened by previously being true. However, if the menu is opened by user click it will ignore `isOpen={false}` and use its internal state. The required `onClose` callback syncs the `myControledBooelan` state to avoid the inconsistency. I think.


```
() => {
  const ref = useRef(null);
  const [isOpen, setIsOpen] = useState(true);
  
  setTimeout(() => {setIsOpen(false)}, 3000)

return (

<div>
  <ADropdown
    ref={ref}
    isOpen={isOpen}
    onClose={() => {
      setIsOpen(false);
    }}
    title="Can be opened by a state prop">
    <AListItem
      onClick={() => {
        console.log(ref.current);
      }}>
      List item
    </AListItem>
    <AListItem
      onClick={() => {
        console.log(ref.current);
      }}>
      List item
    </AListItem>
  </ADropdown>
</div>

);
}
```

brennarvo:
@rwharrisjr @mvelkCisco the dilemma here is that if we allow `isOpen` to fully control the state, then a consumer of `ADropdown` then has to handle all the different closing scenarios (e.g. outside clicks, <kbd>esc</kbd> keypresses, closing after clicking a menu item, etc.).

I think a way we can handle both of these cases is like this:

```typescript
function Example() {

  return (
    <ADropdown
      isOpen={isOpen}
      // `onClose` is called on outside clicks, ESC keypresses, on dropdown item clicks, etc.
      onClose={() => setIsOpen(false)}
    />
  )
}
```

In this example, the `isOpen` prop should always be respected, meaning if I hardcoded `false`, it would *never* open. However — the `onClose` prop would allow a consumer to handle all the different closing scenarios (outside click, <kbd>esc</kbd> keypresses, etc.)

mvelkCisco:
@rwharrisjr @brennarvo I think you both proposing the same thing. I think this makes sense. Sorry, it's a bit late in the day for me :D

brennarvo:
@mvelkCisco I think mine is slightly different. @rwharrisjr is suggesting that even if we pass `isOpen={false}` to the menu, it would still open up if the user re-clicked the menu itself.

To address that with my suggestion, we'd also need an `onOpen` prop:

```tsx
function Example() {

  return (
    <>
      <AButton onClick={() => setIsOpen(true)}>Open</AButton>
      <ADropdown
        isOpen={isOpen}
        // `onOpen` is called whenever the internal `ADropdown` implementation determines it's trying to be opened
        onOpen={() => setIsOpen(true)}
        // `onClose` is called on outside clicks, ESC key, and when a dropdown item is clicked, etc.
        onClose={() => setIsOpen(false)}
      />
    </>
  );
}
```

mvelkCisco:
Right. As a consumer I would expect that the menu does not close when `isOpen={true}`, unless I provide that `onClose={() => setIsOpen(false)}` callback. If you internally call that `onClose` instead of actually closing using your internal state, we can have a good and consistent data flow on both sides. If that's the case, and it's provided as an example in the documentation I think we are good. 

If the internal and external state get out of sync just because the the `onClose` was not provided, that's what can cause unexpected bugs. I don't think all developers will realize that clicking outside the menu will close it for them, and that the states would diverge.

rwharrisjr:
I wonder if it would be overkill to add an error/warning for "if `isOpen` prop exists on props check that `onClose` also exists, else throw error"?

mvelkCisco:
> I wonder if it would be overkill to add an error/warning for "if `isOpen` prop exists on props check that `onClose` also exists, else throw error"?

I think it wouldn't hurt. If the consumer defines `onClick={() => setIsOpen(prev => !prev)}`, they still might not notice an issue.

It would be nice to restrict this by TS.

framework/components/ADropdown/ADropdown.tsx
```diff
@@ -1,42 +1,66 @@
...
+import React, {forwardRef, useCallback, useState} from "react";
 import {useMergeRefs} from "@floating-ui/react";
 
 import AButton from "../AButton";
 import AFloatingMenu from "../AFloatingMenu";
 import useFloatingDropdown from "../AFloatingMenu/useFloatingDropdown";
-import {ADropdownProps} from "./types";
+import type {ADropdownProps} from "./types";
+
+export type MustExist<T, ErrorMessage extends string> = T extends () => void
```
androche, line 10: In the example  `() => void` would actually be the type we would _not_ want it to allow.   So in example it was dont allow any arrays, so it was `T extends any[]` -> pass in array -> ts error.  

I assumed something like `T extends undefined` would error TS - it does when you manually pass in undefined, but not the value from the prop.

BUT none of it works TS wise so I guess it doesnt matter.  This could probably all be deleted to just show the console error. lol  Or at least change it to `undefined` so we can pretend

framework/components/ADropdown/ADropdown.tsx
```diff
@@ -1,42 +1,66 @@
...
+  ? ErrorMessage
+  : T;
+
+function mustExist<T>(arg: T & MustExist<T, "onClose must exist with isOpen">) {
+  if (!arg) {
+    console.error("ADropdown: onClose must exist with isOpen", arg);
```
androche, line 16: We're getting an infinite loop here in other envs:

Screenshot 2024-12-09 at 2 20 37 PM

framework/components/ADropdown/ADropdown.tsx
```diff
@@ -1,42 +1,66 @@
 import PropTypes from "prop-types";
-import React, {forwardRef, useState} from "react";
+import React, {forwardRef, useCallback, useState} from "react";
 import {useMergeRefs} from "@floating-ui/react";
 
 import AButton from "../AButton";
 import AFloatingMenu from "../AFloatingMenu";
 import useFloatingDropdown from "../AFloatingMenu/useFloatingDropdown";
```
androche, line 7: No Idea why styling was impacted 


Screenshot 2024-12-09 at 2 25 57 PM


Verified in IM:
Screenshot 2024-12-09 at 2 15 29 PM

androche (changes requested):

androche (approved):

PR #965, opened 2024-11-19
ADrawer specified responsive sizes

rwharrisjr:
	-	[-] This change has been integrated and tested against 2 out of 3 typescript applications
	-	[-] This change passed unit tests in these applications
	-	[x] The changes are documented in component docs and changelog
	-	[ ] The ESLint plugin has been updated if a new component is added
	-	[x] Test have been added or modified, if appropriate
  
This adds a `size` prop to drawers. The value can either be a string

`type ADrawerSize = "sm" | "md" | "lg" | "xl"`

or an array containing any of those values.

If the value is a string, the drawer width will be set to the corresponding magnetic size width (vs specifying a width in pixels).

If the value is an array, the drawer width will default to (and be no smaller than) the size of the first value in the array, and responsively change size for the remaining values.

---------
Note that this is technically a breaking change, so Typescript validation in certain repositories will fail without a matching update.

Migration step:

`responsiveWidth` is no longer a boolean, but will need to take an array of sizes.

### DIFF
--- a/framework/components/ADrawer/ADrawer.js
+++ b/framework/components/ADrawer/ADrawer.js
@@ -115,8 +115,6 @@ const ADrawer = forwardRef(
       }
     }
 
-    console.log(className);
-
     if (propsClassName) {
       className += ` ${propsClassName}`;
     }
--- a/framework/components/ADrawer/ADrawer.js
+++ b/framework/components/ADrawer/ADrawer.js
@@ -238,7 +238,10 @@ ADrawer.propTypes = {
    * The Drawer renders with the Magnetic defined widths for the Magnetic
    * defined breakpoints.
    */
-  responsiveWidth: PropTypes.bool,
+  responsiveWidth: PropTypes.oneOfType([
+    PropTypes.arrayOf(["sm", "md", "lg", "xl"]),
+    PropTypes.string
+  ]),
 
   /**
    * Specifies the positioning strategy of the drawer. A drawer specified with
--- a/framework/components/ADrawer/ADrawer.js
+++ b/framework/components/ADrawer/ADrawer.js
@@ -112,10 +112,6 @@ const ADrawer = forwardRef(
         else if (typeof responsiveWidth === "string") {
           className += ` a-drawer--size-${responsiveWidth}`;
         }
-        // Otherwise let it have generic responsive breakpoints
-        else {
-          className += " a-drawer--responsive-width";
-        }
       } else if (openHeight) {
         style.height = openHeight;
       }
--- a/framework/components/ADrawer/ADrawer.js
+++ b/framework/components/ADrawer/ADrawer.js
@@ -20,7 +20,6 @@ const ADrawer = forwardRef(
       openHeight,
       openWidth,
       responsiveWidth,
-      size,
       isOpen,
       onClose,
       closeOnOutsideClick,
@@ -98,18 +97,25 @@ const ADrawer = forwardRef(
     if (isOpen && !slim) {
       if (openWidth) {
         style.width = openWidth;
-      } else if (size) {
-        if (Array.isArray(size)) {
+      } else if (responsiveWidth) {
+        // If it's an array, it means we want to make the drawer responsive at
+        // the corresponding breakpoints
+        if (Array.isArray(responsiveWidth)) {
           // Use the first size as the smallest width
-          className += ` a-drawer--size-${size[0]}`;
+          className += ` a-drawer--size-${responsiveWidth[0]}`;
           // Make the drawer responsive to the other sizes
           className +=
-            " " + size.map((s) => `a-drawer--responsive-${s}`).join(" ");
-        } else {
-          className += ` a-drawer--size-${size}`;
+            " " +
+            responsiveWidth.map((s) => `a-drawer--responsive-${s}`).join(" ");
         }
-      } else if (responsiveWidth) {
-        className += " a-drawer--responsive-width";
+        // For a string we want to set a fixed width
+        else if (typeof responsiveWidth === "string") {
+          className += ` a-drawer--size-${responsiveWidth}`;
+        }
+        // Otherwise let it have generic responsive breakpoints
+        else {
+          className += " a-drawer--responsive-width";
+        }
       } else if (openHeight) {
         style.height = openHeight;
       }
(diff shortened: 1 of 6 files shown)


framework/components/ADrawer/types.ts
```diff
@@ -51,6 +53,11 @@ export type ADrawerProps<C extends React.ElementType> = Override<
        * defined breakpoints.
        */
       responsiveWidth?: boolean;
+      /**
+       * Specify the width of the drawer based on width variables. This can accept an
+       * to be responsive to specified sizes
```
jflook10, line 58: Did ln 57 get cut off? "... this can accept an array or string of one of the following: "sm" | "md" | "lg" | "xl"

framework/components/ADrawer/ADrawer.scss
```diff
@@ -156,11 +158,47 @@ $common-transitions: transform $transition-props, width $transition-props,
     width: $drawer-width-default;
 
     @media screen and (min-width: map-get($grid-breakpoints, "xl")) {
-      width: $drawer-width-xl;
+      width: $drawer-width-md;
+    }
+
+    @media screen and (min-width: map-get($grid-breakpoints, "xxl")) {
+      width: $drawer-width-lg;
```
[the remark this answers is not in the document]
rwharrisjr: sm, md, and lg are the sizes for drawers in the magnetic spec - and those match the width variables in ADrawer.scss

The grid breakpoints were defined before we had the magnetic spec to base on - and we have extras outside of what magnetic defines. To keep from breaking things we adjusted close to what we could, but they are off a level (magnetic sm is actually our md)


So there's definitely a mismatch, but I tried to keep things as close to magnetic terms without breaking what already exists.

framework/components/ADrawer/types.ts
```diff
@@ -51,6 +53,11 @@ export type ADrawerProps<C extends React.ElementType> = Override<
        * defined breakpoints.
        */
       responsiveWidth?: boolean;
+      /**
```
androche, line 56: Is there a way we can consolidate `responsiveWidth` boolean (potentially deprecate) with the new size array option?  It seems a bit redundant now.

androche (commented):
Cool! 👍 
One comment below, but also could this be tested in at least one typescript application ahead of time 🙏  (checkbox 1)

framework/components/ADrawer/types.ts
rwharrisjr: I'm inclined to keep both, one generic responsive prop, and one with more specific control (perhaps helpful in the future for smaller viewports)

androche: It just seems like a lot of options which might be confusing for users on the difference between the two.  It seems like they could be consolidated pretty easily.  So maybe if we added `sizes` to take in a boolean `size?: ADrawerSize | ADrawerSize[] | boolean`

Then you could check type isboolean and take out`className += " a-drawer--responsive-width"` from `responsiveWidth` and add the class  to the new conditional statement in the sizes check

rwharrisjr: Refactored to reuse the `responsiveWidth` prop. Leaving it as `responsiveWidth` so this will be a non-breaking change, and we might add in `responsiveHeight` in the future for bottom/top positioned drawers

jflook10 (approved):
expanding the responsiveWidth was a great idea. Looks good 👍

framework/components/ADrawer/types.ts
Graham187K: Since I made the request for `responsiveWidth: boolean` in the first place, I wouldn't be totally against removing `boolean` type as an option, if you did want to consolidate this. I like how this works now.

I don't know of anywhere but Incident Manager using the prop as a boolean right now, so I would be ok with migrating the Incident Drawers to the new array based prop with the next update of this lib.

rwharrisjr: If the array format can handle what you needed I'll take out the boolean. I was just wary of taking it out since it'd technically be breaking, and maybe we still wanted the easy format

framework/components/ADrawer/ADrawer.scss
```diff
@@ -5,8 +5,10 @@ $transition-timing: ease-in-out;
 $transition-props: $transition-duration $transition-timing;
 
 $drawer-width-default: 400px;
-$drawer-width-xl: 576px;
-$drawer-width-xxl: 768px;
+$drawer-width-sm: 400px;
+$drawer-width-md: 576px;
+$drawer-width-lg: 768px;
+$drawer-width-xl: 992px;
```
Graham187K, line 11: i was reading over this again and began to get confused between the drawer width sizes and the breakpoints. 

I think I got it now though, at first I thought `$drawer-width-sm` aligns with `$grid-breakpoints: 'sm'`, but it doesn't. Drawer size and breakpoint width are not the same thing. 

So with that in mind, I think to achieve the same effect as the original `responsiveWidth={true}` props, we would need to provide:

`responsiveWidth={["sm", "lg", "xl"]}`, not `responsiveWidth={["md", "lg", "xl"]}`

It took me a minute to wrap my head around that, but I tested it and it does the job.

framework/components/ADrawer/ADrawer.mdx
```diff
@@ -232,6 +232,73 @@ The drawer can slide-in from the left, right, or bottom of the page.
 }`}
 />
 
+#### Specific Size
```
Graham187K, line 235: I can't leave a comment on unedited code, but the `Responsive Width` section starting on line 190, should be removed now?

framework/components/ADrawer/ADrawer.js
```diff
@@ -98,7 +98,20 @@ const ADrawer = forwardRef(
...
+          className += ` a-drawer--size-${responsiveWidth[0]}`;
+          // Make the drawer responsive to the other sizes
+          className +=
+            " " +
+            responsiveWidth.map((s) => `a-drawer--responsive-${s}`).join(" ");
+        }
+        // For a string we want to set a fixed width
+        else if (typeof responsiveWidth === "string") {
+          className += ` a-drawer--size-${responsiveWidth}`;
+        }
```
Graham187K, line 114: Not sure if this has any effect because I saw that you updated the `types.ts` file but the propType on line 241 `responsiveWidth: PropTypes.bool,` should be changed too?

rwharrisjr: Good catch

Graham187K (approved):

Issue #949, opened 2024-10-29
Update ASlider


androche:
#953 Includes:
- Updates styling and functionality without entire refactor
- Deprecated validation, but left in for backward compatibility
- Adds logic to show more accurate positioning with ticks

Tasks:
- [x] Usage audit (No one is using ASlider - may completely delete deprecations, but could do in the follow up PR)
- [x] Testing in repos (loaded appropriately on IM)
- [x] Functionality - Change calculation to show more accurate positioning rather than just evenly spaced - get ticks as close as possible to label.  (Note* absolute perfect positioning of ticks is slightly off in design as well)

In follow up PR will need
- [x] Convert to typescript (too extensive of changes and might include unintentional breaking due to tightening up)
- [x]  Add unit tests in ts
- [ ]  Add tooltips to labels 

Design
![Screenshot 2024-10-28 at 1 02 30 PM](https://github.com/user-attachments/assets/227edbce-212f-4bbe-a947-15f31d2fe8a3)

PR #968, opened 2024-11-21
A slider b/#949 - Add tests and Typescript

androche:
**Follow up to #949: Add tests (pre conversion) converted to typescript.**

  -	[x] This change has been integrated and tested against 2 out of 3 typescript applications 
  (Not currently used in ts env, however tested against one.  Also tested and verified in another JS application where this component is currently used.
  -	[x] This change passed unit tests in these applications -  N/A
  -	[x] The changes are documented in component docs and changelog
  -	[x] Test have been added or modified, if appropriate
  -    [x] Convert component to tsx, if reasonable
  

I hope not

### DIFF
--- a/cypress/tests/ABadge.cy.tsx
+++ b/cypress/tests/ABadge.cy.tsx
@@ -1,6 +1,5 @@
 import React from "react";
-import {AButton, AIcon, ABadge} from "../../framework";
-import type {ABadgeProps} from "../../framework";
+import {AButton, AIcon, ABadge, type ABadgeProps} from "../../framework";
 import {hexToRgb} from "./helpers";
 
 const ABadgeTest = (props: ABadgeProps) => {
--- a/cypress/tests/ASlider.cy.tsx
+++ b/cypress/tests/ASlider.cy.tsx
@@ -35,22 +35,16 @@ describe("<ASlider />", () => {
     cy.get("[aria-valuenow=75]").should("exist");
   });
 
-  it("should move by 1 step", () => {
+  it("should move by 1 step with arrow key", () => {
+    // Step has a value of 1, so it should move by 1 "tick", the previous above moves by 25, so the handler should move by 25 "ticks"
+    // This is just making sure the step value is working properly and not jumping by several ticks.
     cy.mount(
-      <ASliderTest
-        value={3}
-        step={1}
-        max={10}
-        ticks={[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}
-      />
+      <ASliderTest value={25} step={1} max={100} ticks={[0, 25, 50, 75, 100]} />
     );
 
-    cy.get("[aria-valuenow=3]").should("exist");
-    cy.get("[role=slider]")
-      .focus()
-      .trigger("mousedown", {which: 1})
-      .trigger("mousemove", {clientX: 300, clientY: 0});
-    cy.get("[aria-valuenow=2]").should("exist");
+    cy.get("[aria-valuenow=25]").should("exist");
+    cy.get("[role=slider]").focus().type("{leftArrow}");
+    cy.get("[aria-valuenow=24]").should("exist");
   });
 
   it("should be in varying positions to match number", () => {
--- a/cypress/tests/ASlider.cy.tsx
+++ b/cypress/tests/ASlider.cy.tsx
@@ -1,13 +1,8 @@
-import React, {useState, useEffect} from "react";
-import {ASlider} from "../../framework";
-// import type {ASliderProps} from "../../framework";
-// import { ASliderProps } from "../../framework/components/ASlider/types";
+import React, {useState} from "react";
+import {ASlider, type ASliderProps} from "../../framework";
 
-const ASliderTest = ({value: propsValue = 0, ...rest}): ASliderProps => {
+const ASliderTest = ({value: propsValue = 0, ...rest}: ASliderProps) => {
   const [value, setValue] = useState(propsValue);
-  useEffect(() => {
-    console.log("CAlled");
-  });
   return (
     <ASlider
       data-testid="slider"
--- a/cypress/tests/ASlider.cy.tsx
+++ b/cypress/tests/ASlider.cy.tsx
@@ -1,0 +1,96 @@
+import React, {useState, useEffect} from "react";
+import {ASlider} from "../../framework";
+// import type {ASliderProps} from "../../framework";
+// import { ASliderProps } from "../../framework/components/ASlider/types";
+
+const ASliderTest = ({value: propsValue = 0, ...rest}): ASliderProps => {
+  const [value, setValue] = useState(propsValue);
+  useEffect(() => {
+    console.log("CAlled");
+  });
+  return (
+    <ASlider
+      data-testid="slider"
+      label="Label"
+      value={value}
+      onChange={(v) => setValue(v)}
+      hints={`Value: ${value}`}
+      {...rest}
+    />
+  );
+};
+
+describe("<ASlider />", () => {
+  it("should render with specifications", () => {
+    cy.mount(
+      <ASliderTest
+        value={50}
+        step={25}
+        max={100}
+        ticks={[0, 25, 50, 75, 100]}
+      />
+    );
+
+    cy.get(".a-slider__tick-label").last().should("have.text", "100");
+    cy.get("[aria-valuenow=50]").should("exist");
+    cy.get("[role=slider]")
+      .focus()
+      .trigger("mousedown", {which: 1})
+      .trigger("mousemove", {clientX: 1000, clientY: 0});
+    cy.get("[aria-valuenow=75]").should("exist");
+  });
+
+  it("should move by 1 step", () => {
+    cy.mount(
+      <ASliderTest
+        value={3}
+        step={1}
+        max={10}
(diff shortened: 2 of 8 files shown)


framework/components/ASlider/types.ts
Graham187K, line 1: Suggestion:

Change
`onChange?: (...args: any[]) => unknown;`
to 
`onChange?: (newVal: ASliderValue) => void;` ?

cypress/tests/ABadge.cy.tsx
```diff
@@ -1,6 +1,5 @@
 import React from "react";
-import {AButton, AIcon, ABadge} from "../../framework";
-import type {ABadgeProps} from "../../framework";
+import {AButton, AIcon, ABadge, type ABadgeProps} from "../../framework";
```
Graham187K, line 2: Suggestion:

Add `eslint` rule in future:
```
"@typescript-eslint/consistent-type-imports": [
  "warn",
  {
    prefer: "type-imports"
  }
]
```
Will remind you when importing a type to import as a type and not value. I'd forget to do this all the time without this rule, so it adds some consistency.

cypress/tests/ASlider.cy.tsx
```diff
@@ -0,0 +1,91 @@
...
+        step={15}
+        max={1000}
+        ticks={[0, 10, 90, 250, 500, 750, 800, 840, 1000]}
+      />
+    );
+
+    cy.get(".a-slider__tick-label").should(($els) => {
+      expect($els[0].style.width).to.not.eq($els[1].style.width);
+    });
+  });
```
jflook10, line 69: I'm a little stuck on this test. From looking at the storybook, it seems like the width could be the same depending on the tick and step interval. I'm not sure that this test is needed and if it is, then we should test that the width is an expected value, rather than to.not.eq... maybe cheating and running the test to see what the width is and setting it that way, but that also sounds like a recipe for flakiness.

cypress/tests/ASlider.cy.tsx
```diff
@@ -0,0 +1,91 @@
...
+      />
+    );
+
+    cy.get("[aria-valuenow=3]").should("exist");
+    cy.get("[role=slider]")
+      .focus()
+      .trigger("mousedown", {which: 1})
+      .trigger("mousemove", {clientX: 300, clientY: 0});
+    cy.get("[aria-valuenow=2]").should("exist");
+  });
```
jflook10, line 54: I think this test could be removed since it looks like it overlaps with the first test. Maybe an update to the "it" statement could cover it? 

ex: it should render with specifications and move one step.

framework/components/ASlider/types.ts
```diff
@@ -2,8 +2,8 @@ import {Override} from "../../types";
 import {AFieldBaseProps, AHintsType} from "../AFieldBase/types";
 
 export interface ASliderRules {
-  test?: (...args: any[]) => unknown;
-  level?: string;
+  test?: (...args: any[]) => string;
```
Graham187K, line 5: It might be worth turning this rule on in the long run, if the application using the component is more strictly typed, this might lead them to some confusion as to what is the allowed format of the test rules.

`"@typescript-eslint/no-explicit-any": "error",`

It's a bit of a pain the butt in the short term though to make everything compliant, I'm not 100% what this type should be, maybe it's `(args: ASliderValue[]) => string;`? Or can there be multiple args? The default test in `Aslider.tsx` implies it's one value, so `number` or `number[]` sounds about right?

framework/components/ASlider/ASlider.tsx
```diff
@@ -15,19 +14,20 @@ import AFieldBase from "../AFieldBase";
...
-
-const floatSafeRemainder = (val, step) => {
-  var valDecCount = (val.toString().split(".")[1] || "").length;
-  var stepDecCount = (step.toString().split(".")[1] || "").length;
-  var decCount = valDecCount > stepDecCount ? valDecCount : stepDecCount;
-  var valInt = parseInt(val.toFixed(decCount).replace(".", ""));
-  var stepInt = parseInt(step.toFixed(decCount).replace(".", ""));
+import {ASliderProps, ASliderRules, ASliderValidationState} from "./types";
+
+const floatSafeRemainder = (val: number, step: number) => {
```
Graham187K, line 19: Maybe just add a short comment to explain the fn?

Technically it could be moved into the component and wrapped in `useCallback` but it's not a major deal either way.

framework/components/ASlider/ASlider.tsx
```diff
@@ -42,24 +42,27 @@ const ASlider = forwardRef(
...
       ...rest
     },
     ref
   ) => {
-    const sliderRef = useRef(null);
-    const combinedRef = useCombinedRefs(ref, sliderRef);
+    const sliderRef = useRef<HTMLDivElement>(null);
+    //TODO update when combinedRefs is typed
+    const combinedRef: any = useCombinedRefs(ref, sliderRef);
     const [sliderId] = useState(sliderCounter++);
```
Graham187K, line 53: This is sort of an anti-pattern, though I understand it's an easy way to get a unique id for each instance.

We could also use:

```ts
import {useId} from 'react';
const [sliderId] = useState(useId());
```

framework/components/ASlider/ASlider.tsx
```diff
@@ -179,13 +190,20 @@ const ASlider = forwardRef(
       setIsDown2(false);
     }, [setIsDown2, disabled]);
     const mouseMoveHandler2 = useCallback(
-      (e) => {
+      (e: MouseEvent | TouchEvent) => {
         if (disabled) return;
```
Graham187K, line 194: i did very limited debugging here, but these fn's trigger for each instance of the slider on the page, whenever I trigger a mouse event on 1 of the sliders. I didn't have time to get deep into the cause but there might be a potential perf issue here?

framework/components/ASlider/ASlider.tsx
```diff
@@ -89,19 +92,19 @@ const ASlider = forwardRef(
...
+      const resto = floatSafeRemainder(val, realStep);
       if (resto <= realStep / 2) {
         return val - resto;
       }
 
       return val + realStep - resto;
     };
 
-    const handleChange = (thumb1Pos, thumb2Pos) => {
+    const handleChange = (thumb1Pos: number | null, thumb2Pos: number) => {
```
Graham187K, line 105: Maybe a future consideration to make this impl a bit safer:

When the args are optional are potentially null or undefined, there's potential to call the fn with args in the wrong order, 

Sometimes it's good to make the arg a single object with optional properties.
```
    const handleChange = ({
      position1: thumb1Pos, position2: thumb2Pos
    }: { position1?: number; position2?: number;  }) => {...}
```

This breaks the logic slightly though, because we're saying a value could be `undefined` instead of `null`.
`Math.min()` doesn't like `undefined`, but at the same time, even though it doesn't complain about potential `null` values, it's not strictly giving a "correct" result.

```ts
Math.min(undefined, 4)
NaN // =>> Obvious error, needs to be handled

Math.min(2, 4)
2 // ==> Correct result

Math.min(null, 4)
0 // ==> No error, but it's not clear if this is the intended result, as 0 is not a provided arg.


```

Graham187K (commented):
Nice refactor!

I left some discussion in-line, just to get your opinion. I know the intention of this PR is to do a simple conversion to TypeScript of `ASlider` so I might have gone a bit further than the core intention of this PR. If that's the case I didn't want to make these blockers, maybe just some tech debt items to look into later?

framework/components/ASlider/types.ts
androche: Let's try it!

framework/components/ASlider/ASlider.tsx
androche: I'm honestly not sure what it does myself - definitely some legacy code lol 

I think i'll just add "This appears to help the handler stay within the bounds of the slider"

framework/components/ASlider/ASlider.tsx
androche: Yea true I'm hesitant to mess with it this go around, but @rwharrisjr let me know if you think it would be ok/safe to swap this in?

framework/components/ASlider/ASlider.tsx
androche: Yea in deployed magna we've always had a console warning of ```non-passive event listener to a scroll-blocking 'touchmove' event. Consider marking event handler as 'passive' to make the page more responsive. See``

So probably a good bug to open

framework/components/ASlider/ASlider.tsx
androche: I'll add this that ticket for the above tech/debt!

androche:
@Graham187K Thanks for reviewing this!  I opened https://github.com/cisco-sbg-ui/magna-react/issues/971 for tracking those areas of improvement/more investigation.

cypress/tests/ASlider.cy.tsx
androche: @jflook10 I added some comments, but I also changed the test up a bit to appear more clear about that the handler doesn't have to be on a "visible" tick.  I also changed it to be an arrow key rather than another drag option to make sure we are covering that option as well.  Thanks for picking this out!

Graham187K:
> @Graham187K Thanks for reviewing this! I opened #971 for tracking those areas of improvement/more investigation.

that's excellent, thank you!

Issue #962, opened 2024-11-13
Remove global disabled 


androche:
Remove the global disabled attribute styles and test adding disabled prop to various components to catch side effects if any.

Right now there is a leftover disabled attribute set, however this is setting pointer events to none which impacts areas were someone might want interaction - ex: blocking tooltips from opening when disabled attribute is set on a parent. 

This worked or technically "didnt work" before because we fixed adding the disabled attribute to some components

https://github.com/cisco-sbg-ui/magna-react/blob/main/framework/components/AApp/base/reset.scss#L36

androche:
Resolved in #970

PR #970, opened 2024-11-21
style remove global disabled style

jflook10:

  -	[x] This change has been integrated and tested against 2 out of 3 typescript applications
	-	[x] This change passed unit tests in these applications
	-	[ ] The changes are documented in component docs and changelog
	-	[ ] The ESLint plugin has been updated if a new component is added
	-	[ ] Test have been added or modified, if appropriate
  
  For new components
	-	[ ] The component should be in typescript
	-	[ ] The component should accept a class name as a prop, if appropriate
	-	[ ] The component should accept a forward ref, if appropriate
	-	[ ] Add tests in typescript, cypress/tests/*, if reasonable

If an application was manually applying a disabled attribute or class to a component, the pointer event styles will no longer be applied to that component and its children. Disabled styles and prop still exist for specific components. 

@mixin disabled {
  cursor: not-allowed;
  pointer-events: none;
} 

**Other information**:

Remove the global disabled attribute styles and test adding disabled prop to various components to catch side effects if any.

Right now there is a leftover disabled attribute set, however this is setting pointer events to none which impacts areas were someone might want interaction - ex: blocking tooltips from opening when disabled attribute is set on a parent. 

Imgs show that when manually adding the disabled class or attribute, the child elements no longer inherit the disabled pointer event style. 

disabled mixin: 
@mixin disabled {
  cursor: not-allowed;
  pointer-events: none;
}
 Images show that manually adding the disabled attribute to the component no longer inherit the disable class pointer styles. 
Screenshot 2024-11-21 at 3 39 37 PM

Screenshot 2024-11-21 at 4 33 15 PM

https://github.com/cisco-sbg-ui/magna-react/issues/962

### DIFF
--- a/framework/components/AList/AList.scss
+++ b/framework/components/AList/AList.scss
@@ -14,7 +14,8 @@
     font-weight: unset;
   }
 
-  &--hoverable .a-list-item:not(.a-list-item--two-line) {
+  &--hoverable
+    .a-list-item:not(.a-list-item--two-line):not(.a-list-item--disabled) {
     &:hover,
     a:hover,
     a.a-list-item:hover {
@@ -134,8 +135,9 @@
   }
 
   &-item--disabled:not(.a-list-item--two-line) {
-    color: var(--base-text-disabled);
     background: none;
+    color: var(--base-text-disabled);
+    cursor: not-allowed;
   }
 
   a.a-list-item,
--- a/framework/components/AApp/base/reset.scss
+++ b/framework/components/AApp/base/reset.scss
@@ -29,16 +29,6 @@ body {
     flex: 1 1 auto;
   }
 
-  // -----------------------------------------------------------------------------
-  // Disabled
-  // To disable any element, add a "disabled" attribute or the ".disabled" class
-  // -----------------------------------------------------------------------------
-  &[disabled],
-  & *[disabled],
-  .disabled {
-    @include disabled;
-  }
-
   // Images
   .img,
   .picture {


androche (commented):
Just clicking around and looks like one area we might have to make up for is in `AListItem`

I think this was the very thing we aimed to fix because tooltips wouldn't work once pointer events were taken out, but thinking we need to take out the hover styles that are allowed now.

Before:
Screenshot 2024-12-04 at 10 55 59 AM

After:
Screenshot 2024-12-04 at 10 56 10 AM

androche (changes requested):
Finished clicking around and seems like that's the only place we'll need to make an update to! 👍 👍

framework/components/AList/AList.scss
```diff
@@ -14,7 +14,8 @@
     font-weight: unset;
   }
 
-  &--hoverable .a-list-item:not(.a-list-item--two-line) {
+  &--hoverable
+    .a-list-item:not(.a-list-item--two-line):not(.a-list-item--disabled) {
```
jflook10, line 18: Just added the :not condition for the disabled style. 
I wasn't sure if chaining the :not would create an either/or situation. My understanding is that each not is treated separately. 
Here's what bridgeit says about it: 
_`When you chain multiple :not selectors together, each :not acts independently, and all conditions must be true for the styles to be applied. This effectively creates an "and" condition between the :not selectors.`_

androche: Nice! :) Could we also change it to a default cursor?

jflook10: Yep, changed to cursor not-allowed since it seems like that is the pattern in other disabled styles.

androche: Thanks for checking that - I never noticed that was magnetic's choice! 
Screenshot 2024-12-04 at 3 36 10 PM

androche (approved):

PR #972, opened 2024-12-02
Adds new Entity icons

Graham187K:
Adds the remaining known entity icons to Magna-React.

This is the mapping used in Incidents repo: https://github.com/advthreat/incident-manager/blob/master/src/utils/icons.ts


svg/app-window.svg
androche, line 1: 🥹  aw can we make this our mascot?

svg/circuitry.svg
androche, line 1: Screenshot 2024-12-02 at 3 06 23 PM

Sometimes we have to target more to get it all the same color, but sometimes its intended. I'm not sure what its supposed to look like, but if you did want it to be all one color we can add the name to the list starting line 214 in `AIcon.scss`

svg/seal-question.svg
androche, line 1: Screenshot 2024-12-02 at 3 13 08 PM

These could be added to the "AIcon.scss" file listed above to the dot to pick up theming.

framework/components/AIcon/magnaIcons.js
androche, line 1: According to that map it looks like we might be missing: `folder-notch-open`

androche (commented):
Looks like you got them all! I think I just found one missing and a couple notes on the color differences that might need at least 2 updates.

framework/components/AIcon/magnaIcons.js
Graham187K: Good eye! I should have added this detail: I think that's a bug that I noticed while doing this. I'll submit a fix to Incident manager to change it to `folder-open`.

Here's the Figma for icons which calls for `folder-open` on `file_path`'s: 
https://www.figma.com/design/WzioFqYjVBdO3EXRctNZya/Icons%3A-Observables-and-Assets?node-id=659-191&node-type=frame&m=dev

There's actually no `folder-notch-open` in Phosphor, so I'm not sure how that string came to be in code.

svg/app-window.svg
Graham187K: LOL, now all I can see it a pair of googly eyes when I look at it.

svg/circuitry.svg
Graham187K: Ah thank you, I didn't catch that... I need to enlarge the icons more on my screen to check these things in future.

svg/seal-question.svg
Graham187K: Done, also added `app-window` and the two dots needed it. I think that's all of them.

framework/components/AIcon/magnaIcons.js
androche: Oh weird! :) I wonder what a "notch-open" would have looked like 😄

androche (approved):

PR #975, opened 2024-12-03
docs: PR and versioning process

androche:


Graham187K (approved):
Thanks for dding the Build & publish steps, I was wondering what the correct process was.

PR #977, opened 2024-12-05
Fix: removed unused anchorRef in useFloatingBase

jonathanmcsweet:
Ticket: https://cisco-sbg.atlassian.net/browse/XDR-16386

image

- [ ] This change has been run and tested against at least 2 typescript applications
  
For new components

** NOTE **
I've only verified this in one Typescript app which is our Administration MFE on a development branch in `securex-ui-monorepo` via local linking. If you have another app I can test this against, please let me know.


androche (approved):

rwharrisjr:
Good find

PR #980, opened 2024-12-06
Feat/#978 timeline-updates

### LINKED ISSUE #978: Tidy up activity timeline add numbered list
This will be used in another feature coming up.
A few styles need to be tweaked and add number list option

Design:

Screenshot 2024-12-05 at 3 05 57 PM

androche:
** #978 - Touch up activity timeline and add numbered lists**

- [x] This change has been run and tested against at least 2 typescript applications _**(tested in IM)**_
- [ ] This change passed unit tests in the applications the build was tested against _**(was not able to run unit test against IM due to unrelated failures)**_
- [ ] Convert component to tsx, if reasonable **_(Will revisit in follow up PR)_**
  

**Other information**:

Screenshot 2024-12-05 at 2 59 28 PM

**Numbered Pagination Lists**
CSS will only increment what is on the viewport as seen below in first image.  I was going to leave the option off for paginated lists since they can just add the number and override the styles, but that wouldn't have been very nice. This update also allows a developer to pass `numList` to `AActivityTimelineItem` if using a numbered paginated list to take advantage of same styling.

Before
Screenshot 2024-12-06 at 9 22 59 AM

After
Screenshot 2024-12-06 at 9 18 23 AM

### DIFF
--- a/framework/components/AActivityTimeline/types.ts
+++ b/framework/components/AActivityTimeline/types.ts
@@ -19,7 +19,7 @@ export type AActivityTimelineProps = Override<
     /**
      * Adds numbered timeline and associated positioning to content.
      * If `hasUnorderedItems` is set to `true`, this prop is ignored.
-     * For paginated lists, utilize the accurate number provided in paginated data list rather than `numbered` prop.
+     * For paginated lists, pass the number provided in paginated data to `itemNum` in `AActivityTimelineItem`
      */
     numbered?: boolean;
   }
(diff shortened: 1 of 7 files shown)


framework/components/AActivityTimeline/types.ts
```diff
@@ -94,6 +100,8 @@ export type AActivityTimelineItemProps = Override<
      * for collapsible items (unless it is the last one in the list.)
      */
     withDivider?: boolean;
+    /** Option to pass number to numbered list. Useful for paginated data. */
+    itemNum?: number | string;
```
rwharrisjr, line 104: Good thinking on the number | string, in case a string localized/formatted number is passed in

rwharrisjr (approved):

PR #981, opened 2024-12-12
Feat: turn on noUncheckedIndexedAccess 

jonathanmcsweet:
… to fix compilation errors in securex-ui-monorepo so that we don't have set next.config.js to ignore compilation errors

- [ ] This change has been run and tested against at least 2 typescript applications
- [ ] This change passed unit tests in the applications the build was tested against
  
For new components

**Other information**:

### DIFF
--- a/framework/components/ASlider/ASlider.tsx
+++ b/framework/components/ASlider/ASlider.tsx
@@ -86,11 +86,11 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
       }
     }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
-    const thumb1Position = Array.isArray(value)
+    const thumb1Position = Array.isArray(value) && value?.[0] !== undefined
       ? (value[0] * 100) / (max - min)
       : null;
     const thumb2Position = Array.isArray(value)
-      ? (value[1] * 100) / (max - min)
+      ? ((value?.[1] ?? 1) * 100) / (max - min)
       : (value * 100) / (max - min);
 
     const roundToStep = (val: number) => {
@@ -105,7 +105,7 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
 
     const handleChange = (thumb1Pos: number | null, thumb2Pos: number) => {
       const splitStep = step.toString().split(".");
-      const precision = splitStep.length > 1 ? parseInt(splitStep[1]) : 0;
+      const precision = splitStep.length > 1 ? parseInt(splitStep?.[1] ?? "0") : 0;
 
       if (onChange) {
         if (thumb1Pos !== null) {
@@ -169,7 +169,7 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
           }
 
           const clientX: number =
-            (e as MouseEvent).clientX || (e as TouchEvent).touches[0].clientX;
+            (e as MouseEvent).clientX ?? ((e as TouchEvent)?.touches as TouchList)[0]?.clientX;
 
           const position = Math.min(
             Math.max(0, clientX + offset1),
@@ -201,7 +201,7 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
           }
 
           const clientX: number =
-            (e as MouseEvent).clientX || (e as TouchEvent).touches[0].clientX;
+            (e as MouseEvent).clientX ?? ((e as TouchEvent)?.touches as TouchList)[0]?.clientX;
 
           const position = Math.min(
             Math.max(0, clientX + offset2),
@@ -270,7 +270,7 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
         setError("");
 
         for (let i = 0; i < workingRules.length; i++) {
-          const rule: ASliderRules = workingRules[i];
+          const rule: ASliderRules | undefined = workingRules?.[i];
           if (rule && rule.test) {
             const error = rule.test(testValue);
             const level: ASliderValidationState = rule.level || "danger";
@@ -380,7 +380,9 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
                   if (disabled) return;
                   const target = e.target as HTMLDivElement;
                   setIsDown1(true);
-                  setOffset1(target.offsetLeft - e.touches[0].clientX);
+                  if (e?.touches?.[0]) {
+                    setOffset1(target.offsetLeft - e.touches[0].clientX);
+                  }
                 }}
                 onKeyDown={(e: React.KeyboardEvent<HTMLDivElement>) => {
                   if (disabled) return;
@@ -459,7 +461,9 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
                 if (disabled) return;
                 const target = e.target as HTMLDivElement;
                 setIsDown2(true);
-                setOffset2(target.offsetLeft - e.touches[0].clientX);
+                if (e?.touches?.[0]) {
+                  setOffset1(target.offsetLeft - e.touches[0].clientX);
+                }
               }}
               onKeyDown={(e: React.KeyboardEvent<HTMLDivElement>) => {
                 if (disabled) return;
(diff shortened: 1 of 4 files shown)


framework/components/ASlider/ASlider.tsx
```diff
@@ -86,11 +86,11 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
       }
     }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
-    const thumb1Position = Array.isArray(value)
+    const thumb1Position = Array.isArray(value) && value?.[0] !== undefined
```
jonathanmcsweet, line 89: Screenshot 2024-12-11 at 7 00 26 PM

framework/components/ASlider/ASlider.tsx
```diff
@@ -86,11 +86,11 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
...
       ? (value[0] * 100) / (max - min)
       : null;
     const thumb2Position = Array.isArray(value)
-      ? (value[1] * 100) / (max - min)
+      ? ((value?.[1] ?? 1) * 100) / (max - min)
```
jonathanmcsweet, line 93: Screenshot 2024-12-11 at 7 00 41 PM

framework/components/ASlider/ASlider.tsx
```diff
@@ -105,7 +105,7 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
 
     const handleChange = (thumb1Pos: number | null, thumb2Pos: number) => {
       const splitStep = step.toString().split(".");
-      const precision = splitStep.length > 1 ? parseInt(splitStep[1]) : 0;
+      const precision = splitStep.length > 1 ? parseInt(splitStep?.[1] ?? "0") : 0;
```
jonathanmcsweet, line 108: Screenshot 2024-12-11 at 7 01 22 PM

framework/components/ASlider/ASlider.tsx
```diff
@@ -169,7 +169,7 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
           }
 
           const clientX: number =
-            (e as MouseEvent).clientX || (e as TouchEvent).touches[0].clientX;
+            (e as MouseEvent).clientX ?? ((e as TouchEvent)?.touches as TouchList)[0]?.clientX;
```
jonathanmcsweet, line 172: Screenshot 2024-12-11 at 7 01 53 PM

framework/components/ASlider/ASlider.tsx
```diff
@@ -201,7 +201,7 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
...
```
jonathanmcsweet, line 204: Screenshot 2024-12-11 at 7 02 12 PM

framework/components/ASlider/ASlider.tsx
```diff
@@ -270,7 +270,7 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
         setError("");
 
         for (let i = 0; i < workingRules.length; i++) {
-          const rule: ASliderRules = workingRules[i];
+          const rule: ASliderRules | undefined = workingRules?.[i];
```
jonathanmcsweet, line 273: Screenshot 2024-12-11 at 7 02 34 PM

framework/components/ASlider/ASlider.tsx
```diff
@@ -380,7 +380,9 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
                   if (disabled) return;
                   const target = e.target as HTMLDivElement;
                   setIsDown1(true);
-                  setOffset1(target.offsetLeft - e.touches[0].clientX);
+                  if (e?.touches?.[0]) {
```
jonathanmcsweet, line 383: Screenshot 2024-12-11 at 7 02 57 PM

framework/components/ASlider/ASlider.tsx
```diff
@@ -459,7 +461,9 @@ const ASlider = forwardRef<HTMLDivElement, ASliderProps>(
                 if (disabled) return;
                 const target = e.target as HTMLDivElement;
                 setIsDown2(true);
-                setOffset2(target.offsetLeft - e.touches[0].clientX);
+                if (e?.touches?.[0]) {
```
jonathanmcsweet, line 464: Screenshot 2024-12-11 at 7 03 31 PM

framework/hooks/useOutsideClick/useOutsideClick.d.ts
```diff
@@ -1,5 +1,5 @@
 interface OutsideClickProps {
-  rootRef?: React.RefObject<HTMLElement>;
+  rootRef?: React.RefObject<HTMLElement> | {current: HTMLElement | undefined};
```
jonathanmcsweet, line 2: Screenshot 2024-12-11 at 7 05 27 PM

framework/components/ATriggerBase/ATriggerBase.tsx
```diff
@@ -82,10 +82,7 @@ const ATriggerTooltip = ({
   }, [toggleClose, onClose, interactive, tooltipRef]);
 
   useOutsideClick({
-    isEnabled:
-      ((triggerRef && triggerRef.current) ||
-        (tooltipAnchorRef && tooltipAnchorRef.current)) &&
-      trigger === "click",
+    isEnabled: Boolean(triggerRef?.current) || Boolean(tooltipAnchorRef?.current) && trigger === "click",
```
jonathanmcsweet, line 85: Simplification of the ref.current check. I can use `!!` instead of `Boolean` if you'd prefer

androche:
@jonathanmcsweet just want to check in - it seems like monorepo is type checking magna's internal ts files (and breaking builds?) so we'd definitely want to avoid that because we are converting a lot of files over and there will be continued issues in the future.  (not to mention our env isnt flagging our own files so we'll have to figure out why lol)   Anyway, you all will definitely want to leave out node_modules for type checking in the tsconfig or if its in there figure out why its not ignoring properly:

Here's an example of how we do this in our other repo:
```
  "compilerOptions": {
  "include": ["src/**/*"],
  "exclude": [
    "src/**/*.stories.*",
    "node_modules",
    "./node_modules",
    "./node_modules/*"
  ]
  }
  ```

androche:
Oh and also probably make sure this is set: `"skipLibCheck": true,`

jonathanmcsweet:
> @jonathanmcsweet just want to check in - it seems like monorepo is type checking magna's internal ts files (and breaking builds?) so we'd definitely want to avoid that because we are converting a lot of files over and there will be continued issues in the future. (not to mention our env isnt flagging our own files as diligently so we'll have to figure out why lol) Anyway, you all will definitely want to leave out node_modules for type checking in the tsconfig or if its in there figure out why its not ignoring properly:
> 
> Here's an example of how we do this in our other repo:
> 
> ```
>   "compilerOptions": {
>   "include": ["src/**/*"],
>   "exclude": [
>     "src/**/*.stories.*",
>     "node_modules",
>     "./node_modules",
>     "./node_modules/*"
>   ]
>   }
> ```

> Oh and also probably make sure this is set: `"skipLibCheck": true,`

We have `skipLibCheck: true` and we also skip `node_modules` and the `magna-react/lib/*` folder, but this library has been generally consumed in a manner where the consumer transpiles it in NextJS and that's why we still get these errors.

`transpilePackages: ["@cisco-sbg-ui/magna-react", "@advthreat/xdr-components"]`

Transpiling means we're going to inherit your components as if they were native. Should we stop consuming your library in a way that transpiles them natively? That's ok with me, but regarding this PR, it wouldn't hurt to have increasing strictness for the sake of the library even if we need to move away from transpiling as if it's native.

androche:
@jonathanmcsweet well what's interesting is we export ***-components in the same way - but yes my vote for no transpile!

And yes, we do need to figure out why our env isn't as strict so thanks for helping catch those!  Although, sometimes null checks that more standard and copilot tends to prefer have broken functionality in the past so sometimes we have to opt to alternatives to get the ts happy (like the casting in the e.touches for a-slider).  I didn't see any issues at first glance though, but sharing for reviewers as we just need to make sure we do thorough click testing on these!

framework/components/ATriggerBase/ATriggerBase.tsx
androche: Nice!

androche (approved):
Pulled and tested in another repo.

PR #982, opened 2024-12-13
build/commonjs-as-module

androche:
This makes monorepo stop yelling about us for some reason (I'm guessing it accepts our cjs and moves on).  No idea how this impacts other groups so we should do some thorough testing and run their unit tests


androche:
@rwharrisjr  Closing as I didn't test this prior to the previous commit that went in.  We are instead investigating a cjs build with rollup

PR #983, opened 2024-12-17
XDR-5679 Shared Component: Tabbed Card Stepper

jflook10:
Create a stepper card. 

- [ ] Test have been added or modified, if appropriate (N/A)
- [ ] Convert component to tsx, if reasonable (N/A)
  
For new components
- [x] The component should be in typescript
- [x] Add tests in typescript, cypress/tests/*, if reasonable

N/A

**Other information**:
https://cisco-sbg.atlassian.net/browse/XDR-15679

### DIFF
--- a/cypress/tests/AStepperCard.cy.tsx
+++ b/cypress/tests/AStepperCard.cy.tsx
@@ -1,89 +1,163 @@
 import React, {useState} from "react";
-import {AButton, ACardFooter, ACardItem, AStepperCard, AStepperCardContent} from "../../framework";
+import {
+  AButton,
+  ACardFooter,
+  ACardItem,
+  AStepperCard,
+  AStepperCardContent
+} from "../../framework";
 
 const cardContent = [
-    {title: "Title One", subtitle: "Subtitle", content:<p data-testid="aStepperCard-step-menuitem-1">Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Vestibulum tortor quam, feugiat vitae, ultricies eget, tempor sit amet, ante. </p>}, 
-    {title: "Title Two", subtitle: "Subtitle", content:<p data-testid="aStepperCard-step-menuitem-2">Consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna</p>}, 
-    {title: "Title Three", content:<p data-testid="aStepperCard-step-menuitem-3">Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Vestibulum tortor quam, feugiat vitae, ultricies eget, tempor sit amet, ante. Donec eu libero sit amet quam egestas semper. Aenean ultricies mi vitae est. Mauris placerat eleifend leo.</p>}]
-    
-const steps = [
-    {title: "Long First Stepper Title" },
-    {title: "Second Title"},
-    {title: "Final Title"}
-  ];
+  {
+    title: "Title One",
+    subtitle: "Subtitle",
+    content: (
+      <p data-testid="aStepperCard-step-menuitem-1">
+        Pellentesque habitant morbi tristique senectus et netus et malesuada
+        fames ac turpis egestas. Vestibulum tortor quam, feugiat vitae,
+        ultricies eget, tempor sit amet, ante.{" "}
+      </p>
+    )
+  },
+  {
+    title: "Title Two",
+    subtitle: "Subtitle",
+    content: (
+      <p data-testid="aStepperCard-step-menuitem-2">
+        Consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore
+        et dolore magna
+      </p>
+    )
+  },
+  {
+    title: "Title Three",
+    content: (
+      <p data-testid="aStepperCard-step-menuitem-3">
+        Pellentesque habitant morbi tristique senectus et netus et malesuada
+        fames ac turpis egestas. Vestibulum tortor quam, feugiat vitae,
+        ultricies eget, tempor sit amet, ante. Donec eu libero sit amet quam
+        egestas semper. Aenean ultricies mi vitae est. Mauris placerat eleifend
+        leo.
+      </p>
+    )
+  }
+];
 
-const AStepperCardTest = ({active, setActive, ...rest}) => {
+const steps = [
+  {
+    title: "Long First Stepper Title",
+    visited: false
+  },
+  {
+    title: "Second Title",
+    visited: false
+  },
+  {
+    title: "Final Title",
+    visited: false
+  }
+];
+
+const AStepperCardTest = ({
+  active,
+  setActive,
+  ...rest
+}: {
+  active: number;
+  setActive: (value: number) => void;
+  [key: string]: unknown;
+}) => {
   return (
     <div style={{height: 500}}>
-    <AStepperCard 
-        active={active} 
-        items={steps} 
-        data-testid={"aStepperCard-container"} 
-        stepProps={{visited: false}}>
-    <AStepperCardContent title={cardContent[active - 1].title} 
-        data-testid={"aStepperCardContent-container"}
-        isNextButtonDisabled={active === cardContent.length}
-        nextButtonText={active < cardContent.length ? "Next" : "Submit"}
-        onBack={() => active > 1 && setActive(active - 1)}
-        onCancel={() => {}}
-        onNext={()=> active < steps.length && setActive(active + 1)}
-        {...rest}
-    >
-        {cardContent[active - 1].content}
-    </AStepperCardContent>
-</AStepperCard>
-</div>
+      <AStepperCard
+        active={active}
+        steps={steps}
+        data-testid={"aStepperCard-container"}>
+        <AStepperCardContent
+          title={cardContent[active - 1]?.title}
+          data-testid={"aStepperCardContent-container"}
+          nextButtonProps={{
+            onClick: () => active < steps.length && setActive(active + 1),
+            disabled: active === cardContent.length,
+            "data-testid": "aStepperCardContent-button-next"
+          }}
+          nextButtonText={active < cardContent.length ? "Next" : "Submit"}
+          backButtonProps={{onClick: () => active > 1 && setActive(active - 1), "data-testid": "aStepperCardContent-button-back"}}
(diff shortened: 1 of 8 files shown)


framework/components/AStepperCard/AStepperCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+  ({className: propsClassName = "", active, children, items, stepProps, ...rest}, ref) => {
+    return (
+      <ACardContainer className="a-stepper-card-container pa-0" style={{height: "100%"}} ref={ref} {...rest}>
+        <ARow style={{height: "100%"}} noGutters>
+        <ACol cols="4" className="a-stepper__card--col dark-cool-grey white--text text-center pr-0">
+          <AStepper vertical>
+          {items.map((item: AStepProps, index: number) => {
+          const stepNumber = index + 1;
+          return (
+            <AStep
```
androche, line 23: It looks like step props somehow might not be getting passed through to `AStep` for some props.   For example, It does work for `showIconOnVisited`, but then I added disabled here on 2nd one, but its still appearing as active:
Screenshot 2025-01-15 at 11 06 55 AM

framework/components/AStepperCard/AStepperCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+import AStep from "../AStepper/AStep";
+import {AStepProps} from "../AStepper/types";
+import AStepTitle from "../AStepper/AStepTitle";
+import { AStepperCardProps } from "./types";
+import "./AStepperCard.scss";
+
+const AStepperCard = forwardRef<HTMLDivElement, AStepperCardProps>(
+  ({className: propsClassName = "", active, children, items, stepProps, ...rest}, ref) => {
+    return (
+      <ACardContainer className="a-stepper-card-container pa-0" style={{height: "100%"}} ref={ref} {...rest}>
```
androche, line 16: There appears to be some jumping around when there is less content.  I see in the designs there is some sort of container management - also based on it scrolling when content gets too long.  I wonder if we should add a set height here or no.  Or if we should have users manage it. I cant really decide! My gut says we give a min height or maybe we need to ask design.
Screenshot 2025-01-15 at 11 15 00 AM

Here it is in IM for example: 

Screenshot 2025-01-15 at 11 18 04 AM

framework/components/AStepperCard/AStepperCardContent.tsx
```diff
@@ -0,0 +1,44 @@
...
+  ACardSubTitle,
+  ACardTitle
+} from "../ACard";
+import {ACol} from "../ALayout";
+import AButton from "../AButton";
+import {AStepperCardContentProps} from "./types";
+import {AColProps} from "../ALayout/types";
+import "./AStepperCard.scss";
+
+const AStepperCardContent = forwardRef<HTMLDivElement, AStepperCardContentProps>(
```
androche, line 16: Scrolling looks great!

framework/components/AStepperCard/AStepperCard.mdx
```diff
@@ -0,0 +1,140 @@
...
+    </div>
+   );
+}
+`}
+/>
+
+## Component Props
+
+### StepperCard Props
+
```
androche, line 131: Maybe it would be good to add a link to AStepper props in docs so people can reference AStepper props.

androche (changes requested):
WOW its incredible how clean this is with all the options offered!!! 🥇 🥇   A couple notes, some questions, and adjustments below.  Tested on IM as well!

framework/components/AStepperCard/AStepperCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <ARow style={{height: "100%"}} noGutters>
+        <ACol cols="4" className="a-stepper__card--col dark-cool-grey white--text text-center pr-0">
```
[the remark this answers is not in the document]
rwharrisjr: We do typically try to start with a "root" class like `.a-stepper-card`.

Since the root element is `<ACardContainer>`, I'd set `.a-stepper-card` to it as the root class - since that is in line with the component name.

Then the `ACol` I'd have `.a-stepper-card__col`, since that's a direct child 

That follows the pattern that's (mostly) used elsewhere in magna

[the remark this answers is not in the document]
jflook10: I removed the dark-cool-gray and white--text. Neither were actually working- thanks for catching that :)

framework/components/AStepperCard/AStepperCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+          const stepNumber = index + 1;
+          return (
+            <AStep
+              key={stepNumber}
+              stepNumber={stepNumber}
+              active={active === stepNumber}
+              visited={active > stepNumber}
+              {...stepProps}
+             >
+              <AStepTitle className="AStepTitle">
```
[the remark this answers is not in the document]
jflook10: deleted 👍

framework/components/AStepperCard/AStepperCardContent.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <ACol cols="8" className={`d-flex flex-column ${propsClassName}`} style={{height: "100%"}} ref={ref} {...rest as AColProps}>
+          {title && <ACardHeader className="mx-4 mt-6">
+            <ACardTitle>{title}</ACardTitle>
+          </ACardHeader>
+          }
+          {subtitle && <ACardSubTitle>{subtitle}</ACardSubTitle>}
+          <ACardContent className="overflow-y-scroll ma-4 mb-0 pb-4">{children}</ACardContent>
+            {contentFooter ? contentFooter : <ACardFooter className="a-stepper__card__content--footer justify-space-between py-3">
+<ACardItem attached="right">
+              {  onCancel && <AButton medium tertiary onClick={onCancel} data-testid="aStepperCardContent-button-cancel">
```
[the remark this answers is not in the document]
jflook10: Since the dev using this component can't pass down a data-testid to the button, would we want to add one for them or qa to hook into? There is likely to be one stepper per page so the id would be fairly unique. And they can always create a custom footer if the testid or button functionality isn't working as needed. 

I could also create a props object for the three buttons that included a test id and text value?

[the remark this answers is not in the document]
androche: Yea because they will want to create their own names.  It might be good to do props object as seen in Drawer: https://github.com/cisco-sbg-ui/magna-react/blob/main/framework/components/ADrawer/types.ts#L205

framework/components/AStepperCard/AStepperCard.tsx
jflook10: Yea, min-height seems like a flexible happy medium :)

framework/components/AStepperCard/AStepperCard.tsx
jflook10: Good catch, I have updated the props so now the steps array will accept/spread the step props.

framework/components/AStepperCard/AStepperCardContent.tsx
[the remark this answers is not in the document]
jflook10: updated- only thing I am not comfortable with is the type declaration and the way it is rendering in the docs. It makes me think that the declaration is incorrect. 

ex: backButtonProps?: AButtonProps<C>[]; 
renders as: AButtonProps<"div">[]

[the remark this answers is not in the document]
androche: Ohh, and looks like it's actually worse in Drawer component. lol.  We should probably open a ticket to figure out how to get it more clear for users.
Screenshot 2025-01-17 at 9 09 21 AM

cypress/tests/AStepperCard.cy.tsx
```diff
@@ -0,0 +1,89 @@
...
+    cy.getByDataTestId("aStepperCardContent-button-next").should("have.text", "Submit").and("have.attr", "disabled");
+
+    cy.getByDataTestId("aStepperCardContent-button-back").click();
+    cy.getByDataTestId("aStepperCard-step-menuitem-2").should("exist");
+
+});
+ it("should render with custom footer", () => {
+    const WrapperComponent = () => {
+
+    const ContentFooter = () => <ACardFooter className="ma-3" data-testid="aStepper-contentFooter-custom">
```
rwharrisjr, line 63: Nothing to change now, just for future reference - in some of the other component tests we take the component definition out of the actual test block so it can be reused. Here it's not reused so it's fine. Just mentioning in case you look through other tests.

framework/components/AStepperCard/AStepperCard.scss
```diff
@@ -0,0 +1,20 @@
+@import "../../styles";
+
+.a-stepper-card {
+  height: 100%;
+  min-height: 200px;
+  &__card {
+    &--col {
```
rwharrisjr, line 7: Typically we use `&--<name>` for something like a modifier vs a dom element. Like `&--primary` in `AButton`

framework/components/AStepperCard/AStepperCard.scss
```diff
@@ -0,0 +1,20 @@
...
+      border-right: 1px solid var(--base-border-medium-default);
+      box-shadow: none;
+    }
+  }
+
+  &__content {
+    &--footer {
```
rwharrisjr, line 14: Same as the `&--col` comment above

rwharrisjr (changes requested):

framework/components/AStepperCard/AStepperCard.scss
androche: I probably led her astray there with the `--` in my earlier comment show a nesting technique example

jflook10: Updated to a "__" for both instances. That fits the pattern, right? 

&__col, &__footer

rwharrisjr: Yep, that'll work. 

Just as a note, I did get carried away with that pattern in early days of the Ribbon code, and wound up with `component__b__c__d__e` sorts of classnames. For things with more than 2 nesting levels I started moving the scss definition a bit higher to end up with just one level of `&__`

rwharrisjr (approved):

PR #987, opened 2025-01-06
fix: reduce useEffect call back to 1

### LINKED ISSUE #986: `AApp` does not set the theme when it's wrapped in `<React.StrictMode>`
`AApp` does not set the theme when it's wrapped in `<React.StrictMode>`

https://github.com/user-attachments/assets/0670d6d8-cde0-4818-856d-6c70a45fd8bf

Same minimally reproducible app without `<React.StrictMode>` commented out seems to work as expected:

https://github.com/user-attachments/assets/e0c6ee94-ae39-4b52-a685-dbb4e8cdbee4

```
import React from "react";
import {AApp} from "@cisco-sbg-ui/magna-react";

export default function App() {
  return (
    <React.StrictMode>
      <AApp persistTheme>
        <h1>Some App here</h1>
      </AApp>
    </React.StrictMode>
  );
}
```

androche:
**Resolves #986 - React strict mode changes**

**Problem**

In react StrictMode, `useEffect`'s purposefully run twice to catch flakiness.  In this case, the theme was getting set properly, and then RESET with a bad state conflict of the newly set theme and the rerun of another "initial onload" set theme.

This check just ensures that we resave only when there's been a true onchange event.  This avoids the redundant reloads which cause the ripple effect of mismatching values which ultimately lead to a resetting of "default".

**To recreate issue in x__-components**

1) Go to `withAAppDecorator.jsx` and add:
``` jsx
const withAAppDecorator = (Story, props) => {
  return (
    <React.StrictMode>
      <div style={{minWidth: "95%", height: "70vh"}}>
        <AApp style={{height: "100%"}} persistTheme withNewWrappingContext>
          <ACardContainer
            style={{minHeight: "200px", textAlign: "center", margin: "20px"}}>
            <ACardContent>
              <Story />
            </ACardContent>
          </ACardContainer>
        </AApp>
      </div>
    </React.StrictMode>
  );
};
```

2) Open Application tab and localStorage in dev tools
3) Change `persist-magna-react-theme` to `dusk`.
4) Refresh page and it will flash dark mode, until going back to default.

**To validate fix**

1) In magna `npm run build`, `npm pack`
2) `npm i` this in `x__-components package.json`
3) Repeat steps above and theme `dusk` should stay put

### DIFF
--- a/framework/components/ATheme/ATheme.js
+++ b/framework/components/ATheme/ATheme.js
@@ -96,7 +96,7 @@ const ATheme = forwardRef(
 
     // persist currentTheme on change
     useIsomorphicLayoutEffect(() => {
-      if (shouldUseThemeStorage) {
+      if (shouldUseThemeStorage && currentTheme !== theme) {
         ThemeStorage.saveTheme(currentTheme);
       }
     }, [currentTheme]);


vdineva (approved):
thank you for the quick fix!

rwharrisjr (approved):

PR #991, opened 2025-01-07
Refactor/#978 timeline

androche:
Closes **#978 - Typescript follow up and remove ATimeline**

- [ ] The changes are documented in component docs and changelog - N/A
- [ ] Test have been added or modified, if appropriate - N/A
- [x] Convert component to tsx, if reasonable


framework/components/AActivityTimeline/AActivityTimeline.tsx
```diff
@@ -26,32 +29,14 @@ const AActivityTimeline = forwardRef(
 
     const ListTag = hasUnorderedItems ? "ul" : "ol";
 
-    return (
-      <ListTag className={className} ref={ref} {...rest}>
-        {children}
-      </ListTag>
+    return React.createElement(
```
androche, line 32: TS would not accept any attempt to decipher between UL vs OL list refs, casting, splitting OL and UL to assign them individually, spreading the ref and then casting, splitting the element into two with their own assigned refs...  `createElement` seems to bypass this issue and probably worked out as a more official reacty way to return the element.

framework/components/AActivityTimeline/AActivityTimelineItemTitle.tsx
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  return (
+    <div className={className} ref={ref}>
+      {children}
+    </div>
+  );
+});
+
+AActivityTimelineItemTitle.displayName = "AActivityTimelineItemTitle";
```
jflook10, line 21: Wouldn't the export in line 23 and const declaration automatically apply the same name in devtools? Just curious why we use this.

androche: Not sure tbh - I think it adds to the key/value that the doc generator reads.  @rwharrisjr Do you know more about the origins here?

framework/components/AActivityTimeline/AActivityTimeline.tsx
```diff
@@ -26,32 +29,14 @@ const AActivityTimeline = forwardRef(
...
+      ListTag,
+      {className, ref: ref, ...rest},
```
jflook10, line 34: Do we need to alias ref: ref?

androche: oh no - good catch!

jflook10 (approved):
tested locally in incident manager 👍

PR #992, opened 2025-01-07
Bugfix ADrawer Responsive width array based prop type 

Graham187K:
Bugfix prompted by a discussion in UI dev channel.

image

The existing PropType definition for `ADrawers`s `responsiveWidth` is not quite correct. The Typescript type is correct, but if the build falls back to using proptypes, it may flag the issue.


androche (approved):
🎉

PR #993, opened 2025-01-07
fix(APageAlert): refine prop types

brennarvo:
## Context

When rendering `APageAlert` with React children in a TypeScript project, the following warning is shown:

Screenshot 2025-01-07 at 3 07 02 PM

This seems to be because of a combination of two things:

1. The prop types do not define `children`
2. There are no explicitly defined TypeScript types for the `APageAlert` props

Because of (2), I'm assuming something in this repo defaults to generating the TypeScript types from prop types. However, since `children` is missing in the prop types, the TypeScript type also excludes the `children` prop.

## This PR

* Defines TypeScript types for the `APageAlert` component
* Adds `children` to the `APageAlert` prop types

### DIFF
--- a/framework/components/APageLayout/APageAlert.d.ts
+++ b/framework/components/APageLayout/APageAlert.d.ts
@@ -1,0 +1,9 @@
+import React from "react";
+
+import {type APageAlertProps} from "./types";
+
+declare const APageAlert: (
+  props: React.PropsWithChildren<APageAlertProps>
+) => JSX.Element;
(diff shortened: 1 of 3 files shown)


vdineva (commented):
lgtm

vdineva (approved):

androche (approved):
Nice yep you got it! - Looks like this was accidentally skipped over during the initial TS conversion!

PR #1001, opened 2025-01-13
Add tertiary destructive button state

rwharrisjr:

- [ ] This change has been run and tested against at least 1 typescript application.  (If extensive change, 2 is best)
  
For new components

**Other information**:

This is in the magnetic design, so I think it's fair to add it. Needed for some upcoming work

Screenshot 2025-01-13 at 4 45 04 PM
Screenshot 2025-01-13 at 4 45 15 PM


jflook10 (approved):
in vercel, tested all the states/prop combos including tertiaryAlt destructive and different icons.

PR #1003, opened 2025-01-15
fix: export TS types for ASkeleton

rwharrisjr:

- [ ] This change has been run and tested against at least 1 typescript application.  (If extensive change, 2 is best)
- [ ] This change passed unit tests in the applications the build was tested against
  
For new components

**Other information**:

### DIFF
--- a/framework/components/ASkeleton/index.ts
+++ b/framework/components/ASkeleton/index.ts
@@ -3,4 +3,18 @@ import ASkeletonHeader from "./ASkeletonHeader";
 import ASkeletonBlock from "./ASkeletonBlock";
 import ASkeletonText from "./ASkeletonText";
 
+import type {
+  ASkeletonProps,
+  ASkeletonHeaderProps,
+  ASkeletonBlockProps,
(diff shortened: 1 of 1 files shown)


androche (commented):
This is not necessary and can be closed- All types are exposed in `types/index.ts`

`export * from "../components/ASkeleton/types";`

trent-boyd (approved):
This looks correct

androche (changes requested):
Blocking just in case 😄

rwharrisjr:
Meant to put this as a draft...

rwharrisjr:
Not needed... bad config on the app side

androche:
Just FYI - This wouldn't have been needed anyway - its already exported! 😄

rwharrisjr:
Yeah, I realized that after I put it in. It didn't match the patterns of some of the other components that export it both ways.. so first thought that might help. But was just a config issue :)

androche:
Well types shouldn't be exported in the index files so those should be deleted as well, the components are exported in the index files though.

